PR #189, opened 2014-12-31
update rack to 1.6.0

cheister:
There have been [a lot of changes to rack](https://github.com/rack/rack/compare/1.5.2...1.6.0) between 1.5.2 and 1.6.0 but all of the specs pass at least.  

The main fix I'm interested in is the XSS vulnerability in the default error app that uses Rack::ShowStatus.  The fixed issue is [here](https://github.com/rack/rack/pull/579)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,7 +5,7 @@ GEM
       bundler
       rake
     diff-lcs (1.2.5)
-    rack (1.5.2)
+    rack (1.6.0)
     rake (10.3.2)
     rspec (2.14.1)
       rspec-core (~> 2.14.0)


kares:
we're far from supporting 1.6.0 fully ... our tests passing do not cover Rack's new features. the packed rack is just there as a fallback, it usually never gets used since gem rack is declared in the Gemfile ... what would be more useful is to know (and backport) the XSS details. thanks for the report!

kares:
**1.1.18** has been released (note that this merge is NOT in the release and rack 1.5.2 is vendor bundled for now) addressing the issue with a back-ported `ShowStatus` for out `ErrorApp` ... 1.6.x while should be working will require some more review work before we're fully confident about it. thanks again!

PR #191, opened 2015-01-05
allow uri-like pathes for app.root and gem.path

mkristian:
when the app.root points to uri:classlaoder:// the rack-application can be packed one-to-one into WEB-INF/classes.
similar for setting gem.path to uri:classloader:// allows to embed the gems in WEB-INF/classes.

setting the gem.path is only needed for jruby-1.7.x and app.root only works with the 1.7.19+. all this allows a simplified packaging: https://github.com/mkristian/jruby-pack#packing-jarwar-files-of-ruby-application
and bypasses all the need of real-path.

### DIFF
--- a/src/main/java/org/jruby/rack/DefaultRackApplicationFactory.java
+++ b/src/main/java/org/jruby/rack/DefaultRackApplicationFactory.java
@@ -7,8 +7,10 @@
 
 package org.jruby.rack;
 
-import java.io.IOException;
+
 import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


kares:
@mkristian interesting thanks, while I appreciate the new direction this is getting super-ugly **uri:** special handling now leaking on 3 places ... will give it some more thoughts later.

mkristian:
@kares I did reduce the patch to something generic ;)

kares:
looks better - thanks! but I'm not sure if it will pass the tests (will get 1.1-stable green just bare with me), the `real_path` code is a bit more responsible than it should be - I think it is expected to return nil (as getRealPath does) for resources that do not exists, otherwise things might not work as expected (e.g. Rails 2/3/4 detection). 

what I'm thinking as an alternative (please let me know what you think) ... is introducing another layout since when all is under `WEB-INF/classes`. it's needs a bit different packaging than Warbler does right? thus configuring a layout sounds appropriate. let me know if I did miss/misunderstood something ...

mkristian:
yes, indeed a new layout will be the better way. even with the current
"patch" it needs quite some configuration in the web.xml to fit the
"alternative" packaging. so a new layout could be much better way to go.
thanx for the input - much appreciated.

mkristian:
@kares this a new layout - so there are no side effects for existing apps.

any idea how to get rid off:
```xml
 <context-param>
    <param-name>rackup</param-name>
    <param-value>eval File.read( 'uri:classloader://config.ru' )</param-value>
 </context-param>
```
when running jetty or tomcat without packing a war then there is no WEB-INF/classes - the classes directory is somewhere else and jruby-rack does not find the config.ru !?

kares:
looks good to me (assuming it works with the class-loader) that WebInf in the name evokes smt "special" about the layout ... have you considered something like `ClassPathLayout` (I know naming is the hard) ?

that `eval` is a uglish - probably we do not use the whole cp to locate the **config.ru** script while booting (it sure happens before the layout is set and used) ... might look into that in detail later if needed.

mkristian:
so I did change the classname ;)

and added the search on context-classloader as last place to search for config.ru

mkristian:
@kares so from my side: I am excited. well, if this goes into 1.1-stable then I would add a small IT on jruby itself for it - it makes it easy to bundle a rack application with maven.

kares:
@mkristian looks good thanks, I'll sure get this into **1.1.18** (will do some testing myself) ... I'll just need to figure out something around the `ErrorApp` - should be released next week. thanks!

kares:
on 1.1-stable with a cherry on top, thanks!

PR #194, opened 2015-04-07
JRuby::Rack::ErrorApp response body is ignored

yousuketto:
When env["rack.showstatus.detail"] is true, JRuby::Rack::ErrorApp response body is overwritten by JRuby::Rack::ErrorApp::ShowStatus.
https://github.com/jruby/jruby-rack/blob/1.1.18/src/main/ruby/jruby/rack/error_app/show_status.rb#L23-L31
Currently, env["rack.showstatus.detail"] will always be true.
https://github.com/jruby/jruby-rack/blob/1.1.18/src/main/ruby/jruby/rack/error_app.rb#L59

### DIFF
--- a/src/main/ruby/jruby/rack/error_app.rb
+++ b/src/main/ruby/jruby/rack/error_app.rb
@@ -64,6 +64,7 @@ module JRuby
       end
 
       def serve(code, path, env)
+        env['rack.showstatus.detail'] = nil
         last_modified = File.mtime(path).httpdate
         return [ 304, {}, [] ] if env['HTTP_IF_MODIFIED_SINCE'] == last_modified
 
--- a/src/spec/ruby/jruby/rack/error_app_spec.rb
+++ b/src/spec/ruby/jruby/rack/error_app_spec.rb
@@ -59,6 +59,7 @@ describe 'JRuby::Rack::ErrorApp' do
       expect( body = response[2] ).to be_a JRuby::Rack::ErrorApp::FileBody
       content = ''; body.each { |chunk| content << chunk }
       expect( content ).to eql '-503-'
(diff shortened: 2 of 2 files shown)


kares:
Hey Yousuke, I'm not really sure I follow ... we're meant to override the body (and thus not use what `@app.call` returned) since we're handing an error and the `ErrorApp` is supposed to return an errror specific body.

Some of your hints are useful and I will use them to tune the logic but the commit as is can not get in.
Maybe explain your use-case if it's no too "weird" :) we can maybe do something about it ...

yousuketto:
#### *About this pull reqest*
Response body was not overridden in v1.1.10, 
but response body is overridden in v1.1.18.
Therefore, I thought it is a bug.
However, it means that specification was changed?


#### *About my problem...*
Thanks to your kind instruction, I understood my problem can be solved by editing my web.xml.
So, It is no longer a big problem.
```xml
<!-- web.xml -->
  <context-param>
    <param-name>jruby.rack.error.app</param-name>
    <param-value>require 'jruby/rack/error_app'; run JRuby::Rack::ErrorApp.new</param-value>
  </context-param>
```

kares:
good work-around on next version I was hoping to rewrite the `ErrorApp` a bit (we'll see when and if that happens), for now I tuned the related code [slightly](https://github.com/jruby/jruby-rack/commit/e3691462bb3d3ac34e6c650265eeabf5f5cca040#diff-0e827a577f6aaa96257867b8f3ef95baL60) so that there's more user-control ... now one can set `env['rack.showstatus.detail'] = false` and if the response has content and is not an error it won't "override the body" ... thanks for the report.

yousuketto:
thanks :thumbsup:

PR #199, opened 2015-10-23
removed use of deprecated method in Rakefile

jhstatewide:
spec.installation_path has been deprecated since 2011


kares:
thanks, has been handled previously https://github.com/jruby/jruby-rack/commit/ee6f1646b9584f806744535111504876290ce8f7

PR #206, opened 2016-05-09
When handling OPTION calls ignore both 'Date' and 'Allow' headers to …

### LINKED ISSUE #205: OPTIONS calls don't get passed to application in Jetty
This is the same problem as in #153, but this time with Jetty instead of Tomcat. Jetty's implementation of `DefaultServlet#doOptions` sets both 'Allow' and 'Date' and therefore slips through the fix in #153.
1. install [rack-cors](https://github.com/cyu/rack-cors) gem
2. configure policy, ensure it runs with webrick
3. deploy app in jetty-runner / war (using jruby-rack)
4. Only Allow and Date are set, rack-cors will not be called

retoo:
…decide if call has been handeled

If not, no OPTION calls will be passed down to the rack app due to the fact
that some implementations of DefaultServlet (i.e. Jetty and Tomcat) already sets
these fields. Fixes #205

### DIFF
--- a/src/main/java/org/jruby/rack/servlet/ResponseCapture.java
+++ b/src/main/java/org/jruby/rack/servlet/ResponseCapture.java
@@ -12,8 +12,11 @@ import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


kares:
merged manually onto 1.1-stable for (1.1.21), thanks

retoo:
@kares are there any plans to release a new version anytime soon?

kares:
1.1.21 was released a few months back, not working for you?

retoo:
Sorry, I was blind:

release

Issue #211, opened 2017-08-11
Java servlet store with Rails 5.1 and Tomcat 7


HarlemSquirrel:
I am trying to set up my rails app to use `:java_servlet_store` in place of `:cookie_store` for my session store. 

Rails throws `undefined method 'get_header' for Rack::Handler::Servlet::DefaultEnv` being thrown when adding the following configuration:

```ruby
if defined?($servlet_context)
  require 'action_controller/session/java_servlet_store'
  # tell rails to use the java container's session store
  Rails.application.config.session_store :java_servlet_store
else
  Rails.application.config.session_store :cookie_store, key: '_servicesdirectory_session'
end
```

Full backtrace:
```
Aug 11, 2017 7:47:14 PM org.apache.catalina.core.ApplicationContext log
INFO: I, [2017-08-11T19:47:14.637448 #1]  INFO -- : [150177b8-e353-437b-b893-a951d7029046] Started GET "/jruby-rails-tomcat-sample/" for 172.19.0.1 at 2017-08-11 19:47:14 +0000

Aug 11, 2017 7:47:14 PM org.apache.catalina.core.ApplicationContext log
INFO: I, [2017-08-11T19:47:14.736209 #1]  INFO -- : [150177b8-e353-437b-b893-a951d7029046] Processing by MessagesController#index as HTML

Aug 11, 2017 7:47:14 PM org.apache.catalina.core.ApplicationContext log
INFO: I, [2017-08-11T19:47:14.826652 #1]  INFO -- : [150177b8-e353-437b-b893-a951d7029046] Completed 500 Internal Server Error in 80ms

Aug 11, 2017 7:47:14 PM org.apache.catalina.core.ApplicationContext log
INFO: F, [2017-08-11T19:47:14.844744 #1] FATAL -- : [150177b8-e353-437b-b893-a951d7029046]   

Aug 11, 2017 7:47:14 PM org.apache.catalina.core.ApplicationContext log
INFO: F, [2017-08-11T19:47:14.846148 #1] FATAL -- : [150177b8-e353-437b-b893-a951d7029046] NoMethodError (undefined method `get_header' for #<Rack::Handler::Servlet::DefaultEnv:0x39f27856>):

Aug 11, 2017 7:47:14 PM org.apache.catalina.core.ApplicationContext log
INFO: F, [2017-08-11T19:47:14.846964 #1] FATAL -- : [150177b8-e353-437b-b893-a951d7029046]   

[...]

PR #212, opened 2017-08-15
Add Rack::Handler::Servlet::DefaultEnv#get_header

HarlemSquirrel:
Fixes https://github.com/jruby/jruby-rack/issues/211

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ group :default do
   if rack_version = ENV['RACK_VERSION']
     gem 'rack', rack_version
   else
-    gem 'rack', '~> 1.5.2'
+    gem 'rack', '~> 2.0.3'
   end
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ group :default do
   if rack_version = ENV['RACK_VERSION']
     gem 'rack', rack_version
   else
-    gem 'rack', '~> 2.0.3'
+    gem 'rack', '~> 1.5.2'
   end
 end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,7 @@ GEM
       rake
     diff-lcs (1.2.5)
     jruby-openssl (0.9.21-java)
-    rack (1.5.5)
+    rack (2.0.3)
     rake (10.4.2)
     rspec (2.14.1)
       rspec-core (~> 2.14.0)
@@ -23,9 +23,9 @@ PLATFORMS
 DEPENDENCIES
   appraisal
   jruby-openssl (~> 0.9.20)
-  rack (~> 1.5.2)
+  rack (~> 2.0.3)
   rake (~> 10.4.2)
   rspec (~> 2.14.1)
 
 BUNDLED WITH
-   1.14.6
+   1.15.3
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,7 @@ GEM
       rake
     diff-lcs (1.2.5)
     jruby-openssl (0.9.21-java)
-    rack (2.0.3)
+    rack (1.5.5)
     rake (10.4.2)
     rspec (2.14.1)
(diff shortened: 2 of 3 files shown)


kares:
thanks for getting sessions working, hopefully the rest of the code-base is compatible with Rails 5.x as well :)
guess its done as is or are you planning on adding tests?

HarlemSquirrel:
I just bumped the rack version to 2.0.3 and everything is passing! I think that may be the best way to test it, no? 

Without my changes there are 3 failures:

```
Failures:

  1) Rack::Handler::Servlet (default) env behaves like env sets attributes with false/null values
     Failure/Error: expect( env['org.false'] ).to be false
       
       expected #<FalseClass:0> => false
            got #<String:5364> => ""
       
       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     Shared Example Group: "env" called from ./src/spec/ruby/rack/handler/servlet_spec.rb:819
     # ./src/spec/ruby/rack/handler/servlet_spec.rb:376:in `block in (root)'
     # org/jruby/RubyBasicObject.java:1691:in `instance_eval'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'

  2) Rack::Handler::Servlet lazy env behaves like env sets attributes with false/null values
     Failure/Error: expect( env['org.false'] ).to be false
       
       expected #<FalseClass:0> => false
            got #<String:5624> => ""
       
       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     Shared Example Group: "env" called from ./src/spec/ruby/rack/handler/servlet_spec.rb:835
     # ./src/spec/ruby/rack/handler/servlet_spec.rb:376:in `block in (root)'
     # org/jruby/RubyBasicObject.java:1691:in `instance_eval'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'

  3) Rack::Handler::Servlet servlet-env behaves like env sets attributes with false/null values
     Failure/Error: expect( env['org.false'] ).to be false
       
       expected #<FalseClass:0> => false
            got #<String:5886> => ""
       
       Compared using equal?, which compares object identity,
       but expected and actual are not the same object. Use
       `expect(actual).to eq(expected)` if you don't care about
       object identity in this example.
     Shared Example Group: "env" called from ./src/spec/ruby/rack/handler/servlet_spec.rb:1000
     # ./src/spec/ruby/rack/handler/servlet_spec.rb:376:in `block in (root)'
     # org/jruby/RubyBasicObject.java:1691:in `instance_eval'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'
     # org/jruby/RubyArray.java:2486:in `map'

Finished in 44.82 seconds
581 examples, 3 failures, 2 pending
```

kares:
OK, thanks. Yes the bump is a good preliminary test. 
Wasn't sure how well your changes are tested esp. in relation to the session store ...

However, for 1.1.x line we should not update the (bundled) rack version, unsure of Rack 2.0's details but tests are likely no where near testing it our for real on a servlet container. So the version bump should get reverted.

HarlemSquirrel:
That makes sense. I've added a revert commit. All seems to be well with Rails 5.1.3 and JRuby 9.1.12.0 just so you know!

kares:
looking good, thank you! 1.1 release hopefully next week or so ...

PR #214, opened 2018-01-09
Add error message to log when runtime error is handled

### LINKED ISSUE #213: Runtime error not logging specific message
I recently upgraded a rails app and was deploying it to our `staging` environment with warbler. Every time I made a web request to it though, I kept hitting an unknown error. The only message I saw in my tomcat logs was [`resetting rack response due exception`](https://github.com/jruby/jruby-rack/blob/710c1a4cbde684f1d0bec5907ba3071d51de0a96/src/main/java/org/jruby/rack/AbstractRackDispatcher.java#L52).

After getting some help from some fine folks in the jruby IRC channel, they suggested to make an issue to see if there's enough info being logged. Perhaps I just missed the error message or have misconfigured something? They mentioned that the web response being sent back should have included something about the error but I didn't see anything. I'm guessing it may be because I was running things in `staging` and that Rails just redirects to my `500.html` error page instead of showing any info on the error.

I included the exception message in [this log]((https://github.com/jruby/jruby-rack/blob/710c1a4cbde684f1d0bec5907ba3071d51de0a96/src/main/java/org/jruby/rack/AbstractRackDispatcher.java#L52) and hacked together a build to test for myself. That helped me solve my own issue but it probably doesn't hurt to have this in place for others.

daneshd:
Fixes #213 

~~It may be worth including a backtrace in the message but I opted for a minimal change.~~ (added them)

### DIFF
--- a/src/main/java/org/jruby/rack/AbstractRackDispatcher.java
+++ b/src/main/java/org/jruby/rack/AbstractRackDispatcher.java
@@ -46,10 +46,10 @@ public abstract class AbstractRackDispatcher implements RackDispatcher {
             final RackResponseEnvironment response) throws IOException {
         
         if ( response.isCommitted() ) {
-            context.log(RackLogger.ERROR, "couldn't handle exception "+ e +" : response is committed");
+            context.log(RackLogger.ERROR, "couldn't handle exception so response is committed:", e);
             return;
         }
-        context.log(RackLogger.ERROR, "resetting rack response due to exception:" + e);
+        context.log(RackLogger.ERROR, "resetting rack response due to exception:", e);
         response.reset();
 
         afterException(request, e, response);
--- a/src/main/java/org/jruby/rack/AbstractRackDispatcher.java
+++ b/src/main/java/org/jruby/rack/AbstractRackDispatcher.java
@@ -49,7 +49,7 @@ public abstract class AbstractRackDispatcher implements RackDispatcher {
             context.log(RackLogger.ERROR, "couldn't handle exception "+ e +" : response is committed");
             return;
         }
-        context.log(RackLogger.DEBUG, "resetting rack response due exception");
+        context.log(RackLogger.ERROR, "resetting rack response due to exception:" + e);
         response.reset();
 
         afterException(request, e, response);


kares:
think there's a reason why its not printed here - the abstract `handleException` wasn't meant to be the "main" exception logging piece, although I do not mind the first commit - but don't like the second "related" change : 
```
context.log(RackLogger.ERROR, "couldn't handle exception "+ e +" : response is committed");
```
... is meant to signal an unusual state - someone down the pipeline already commiting the response
the new message is more confusing + the stack trace should have been logged in that case elsewhere.

kares:
> I recently upgraded a rails app and was deploying it to our staging environment with warbler. Every time I made a web request to it though, I kept hitting an unknown error. The only message I saw in my tomcat logs was resetting rack response due exception.

could we have details around the issue - maybe include the exception/stack-trace that you hunted down?
as it must be a quite specific case since most of the time an uncaught (Ruby) exception is wrapped and logged properly when it reaches out of the runtime.

daneshd:
>think there's a reason why its not printed here - the abstract handleException wasn't meant to be the "main" exception logging piece, although I do not mind the first commit

Right, we figured it may be logged/sent somewhere else. I wasn't seeing anything on my end though, maybe that's just my config issue?

>but don't like the second "related" change

Fair enough, I thought I may as well add that in while I was there but I'm happy to remove it. Are you okay with the stack trace on line 52 or would you prefer if it was just the exception message as well?

>could we have details around the issue 

Absolutely! I was deploying a `staging` environment but my `config/secrets.yml` file did not a `secret_key_base` set up for that env. 

Before my change, the only message I saw in the logs was `resetting rack response due exception`. With my changes, I saw this:

```ERROR: resetting rack response due to exception:
org.jruby.exceptions.RaiseException: (RuntimeError) Missing `secret_token` and `secret_key_base` for 'staging' environment, set these values in `config/secrets.yml`
        at var.lib.tomcat7.webapps.ROOT.WEB_minus_INF.gems.gems.railties_minus_4_dot_2_dot_10.lib.rails.application.validate_secret_key_config!(/var/lib/tomcat7/webapps/ROOT/WEB-INF/gems/gems/railties-4.2.10/lib/rails/application.rb:534)
        at var.lib.tomcat7.webapps.ROOT.WEB_minus_INF.gems.gems.railties_minus_4_dot_2_dot_10.lib.rails.application.env_config(/var/lib/tomcat7/webapps/ROOT/WEB-INF/gems/gems/railties-4.2.10/lib/rails/application.rb:246)
        at var.lib.tomcat7.webapps.ROOT.WEB_minus_INF.gems.gems.railties_minus_4_dot_2_dot_10.lib.rails.engine.call(/var/lib/tomcat7/webapps/ROOT/WEB-INF/gems/gems/railties-4.2.10/lib/rails/engine.rb:514)
        at var.lib.tomcat7.webapps.ROOT.WEB_minus_INF.gems.gems.railties_minus_4_dot_2_dot_10.lib.rails.application.call(/var/lib/tomcat7/webapps/ROOT/WEB-INF/gems/gems/railties-4.2.10/lib/rails/application.rb:165)
        at uri_3a_classloader_3a_.rack.handler.servlet.call(uri:classloader:/rack/handler/servlet.rb:22)```

It was looking like this with just my first commit:

```ERROR: resetting rack response due to exception:
org.jruby.exceptions.RaiseException: (RuntimeError) Missing `secret_token` and `secret_key_base` for 'staging' environment, set these values in `config/secrets.yml````

kares:
pushed https://github.com/jruby/jruby-rack/commit/4ed7a4e4f353d7fcf0cf33f9cba9b9b3c45d7dc2 with some tuning https://github.com/jruby/jruby-rack/commit/5854db81cba68ec828908ca1cb646b896faeec64 (explained in the message). that should resolve.

PR #250, opened 2024-05-23
Run specs in CI

headius:
These are not all green but it's better to have them running than not.


chadlwilson:
@headius I think superseded by #262 (although I did not cherry pick from here I believe as don't think I noticed this PR so you might want to check if there is anything here that is a better approach than mine?)

chadlwilson:
@kares would you mind doing some cleanup here and closing this redundant PR? 🙏

PR #252, opened 2024-05-31
Remove usage of Rack::Utils.bytesize

jlahtinen:
See also https://github.com/rack/rack/commit/7b5820f8de7f9dffeb03d86588bba13fb0e91df1

### DIFF
--- a/src/main/ruby/jruby/rack/error_app.rb
+++ b/src/main/ruby/jruby/rack/error_app.rb
@@ -70,14 +70,13 @@ module JRuby
         headers = { 'Last-Modified' => last_modified }
         DEFAULT_HEADERS.each { |field, content| headers[field] = content }
         ext = File.extname(path)
+        size = File.size?(path)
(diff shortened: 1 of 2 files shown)


enebo:
LOL.  I guess CI is still not quite there. 

@jlahtinen Can you also fix: src/main/ruby/jruby/rack/error_app.rb for the same use of bytesite.  That should eliminate any use of that method.

headius:
CI looks fine, it's just that the dependency upload part (for GitHub security checking) is not working (and disabled on master for that reason).

I'm going to merge this for 1.2.3 and we can chat with @kares if a similar fix and update to newer Rack would be appropriate for 1.1.x.

headius:
Correction: disabled **except for master** because PRs should not upload dependencies.

leduytung:
Have you released this fix yet?
The latest version is still 1.2.2(June 10, 2024).

Issue #247, opened 2024-05-23
Broken multipart data


headius:
Original issue is at https://github.com/ruby/strscan/issues/92.

The problem manifested as an ArgumentError "invalid byte sequence in UTF-8" when trying to upload a PDF via Tomcat and jruby-rack to a simple application. Thanks to the simple reproduction by @jlahtinen I was able to narrow this down to a jruby-rack bug.

The code in jruby-rack has not been updated since JRuby 1.7, and as a result it was using some unsafe methods for concatenating bytes onto an existing string. Specifically, it called an old form of `RubyString#cat` that did not consider encoding or code range, and when it appended encoded or binary data onto an existing buffer it could break that buffer for any character operations.

Fix coming.

headius:
@jlahtinen The fix is in, but I have additional work to get this building again. Bear with me! We'll have a new release of this out soon (almost a decade after the previous release) and no doubt there will be some complications. Work with me and we'll get it sorted out!

jlahtinen:
@headius Thank you. I appreciate the progress being made and I understand that this takes as long as it takes :).

headius:
I'm on the cusp of releasing this thing and all that stands in my way is getting authentication for pushing the maven artifact 🙄

headius:
@jlahtinen Ok it's all released! jruby-rack 1.2.0 is released as a gem and maven artifact. Let's see if it works for you and shake out any remaining issues!

jlahtinen:
Currently our app throws exception in tomcat when started

```
2024-05-29T05:15:41.244Z SEVERE org.apache.catalina.core.ApplicationContext log: initialization failed^_
org.jruby.rack.RackInitializationException: load error: /usr/share/tomcat9/sikke_apps/ROOT/WEB-INF/config/environment -- java.lang.NullPointerException: null
        from org/jruby/RubyKernel.java:1184:in `require'
        from uri:classloader:/jruby/rack/rails/environment3.rb:25:in `load_environment'
        from uri:classloader:/jruby/rack/rails_booter.rb:83:in `load_environment'
        from <script>:1:in `<main>'

        at org.jruby.rack.RackInitializationException.wrap(RackInitializationException.java:30)
        at org.jruby.rack.RackApplicationFactoryDecorator.init(RackApplicationFactoryDecorator.java:107)
        at org.jruby.rack.RackServletContextListener.contextInitialized(RackServletContextListener.java:50)
[...]

PR #255, opened 2024-06-06
Coderange cat for 9.3

headius:
This is equivalent to the fix in #248 but using MethodHandle to call an appropriate version for the host JRuby.

It is also targeted at 1.1.x to have a compatible release of that version of jruby-rack without introducing larger changes.

See https://github.com/jruby/jruby-rack/issues/247#issuecomment-2151517962


kares (approved):
LGTM

PR #258, opened 2024-06-11
Correct javax.activation scope

chadlwilson:
This only seems to be used within tests at https://github.com/jruby/jruby-rack/blob/a965a33a23a484371e81a01048714106acda6230/src/spec/java/org/jruby/rack/mock/MockServletContext.java#L466-L468 so note needed as a new runtime dependency (especially as it's a very old API version)?


headius:
Seems appropriate to me!

PR #260, opened 2024-06-19
Change JrubyRackLogger to Logger

jlahtinen:
I am not sure if Logger is always loaded or not. With rails app it seems it is.

### DIFF
--- a/src/main/ruby/jruby/rack.rb
+++ b/src/main/ruby/jruby/rack.rb
@@ -53,14 +53,13 @@ module JRuby
       end
 
       @@logger = nil
-      # Returns a {Logger} instance that uses the {#context} as a logger.
-      def logger; @@logger ||= Logger.new(context!) end
+      # Returns a {Logger} instance. Wraps ServletLog with context.
+      def logger; @@logger ||= ::Logger.new(logdev) end
       # @private
       def logger=(logger); @@logger = logger end
 
       private
 
-      # @deprecated Mostly for compatibility - not used anymore.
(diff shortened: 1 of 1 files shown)


jlahtinen:
I can not say exactly why this initializes Rails.logger in a way it works with requests that has a session management and silence_logger is called.

For some reason silence_logger is not added to JRuby::Rack::Logger but it is added to ::Logger

headius:
@kares Got a sec to look this over? This was a necessary change to get @jlahtinen's app running with jruby-rack 1.2.x.

kares:
So this kind of restores the 1.1 behavior. Which is not the intent here, the idea was to ship with a custom JRuby-Rack logger impl by default. The reason being having better Ruby logging -> Java logging backend integration. Which was not that good in 1.1.x I think e.g. mapping ruby logging level properly to the Java one (would need to look deeper).

For the record, users could use their custom `config.logger` configuration in Rails, which is respected (and this code won't be used if they set that in Rails).

> For some reason silence_logger is not added to `JRuby::Rack::Logger` but it is added to `::Logger`

Would be nice to have a proper fix, the methods missing but required by recent Rails should be added to the `JRuby::Rack::Logger` impl

jlahtinen:
Thanks for the update. I try to investigate this further.

jlahtinen:
This problem rose because the usage of activerecord-session_store version that had its own silence_logger implementation. In our case it was activerecord-session_store 1.1.3.

This problem is gone when using latest activerecord-session_store. I think that is ok and this can be closed.

PR #253, opened 2024-06-04
Merge 1.1 into master

headius:
This is an attempt to merge in all changes from 1.1-stable into master for the next 1.2.x release.

The 1.1-stable branch diverged almost ten years ago, and only had one merge back to master in early 2015. Since then most maintenance and bug-fixing has occurred on the branch. This PR attempts to bring in all the changes for 1.1 and resolve the diff between the two branches so we can move forward with 1.2 releases.

PR #261, opened 2024-06-22
Re-enable specs to run via JRuby-provided bundler

chadlwilson:
This is work-in-progress to get the tests running/passing again against latest JRuby.

**Getting tests/specs passing again**
- [x] Resolve the issue in https://github.com/jruby/jruby-rack/commit/477502ce032107f46110d192b10adc3869d3b6d1
  - Root problem seems to be that running the Rake `ruby` task ended up trying to execute an incorrect Ruby binary via `ENV['RUBY]` when running inside JRuby, and also was trying to run system ruby when run with `FULL_BIN_PATH`, and thus dependent on the environment around it (rather than how Maven was configured with jruby jars).   
- [x] Resolve errors/warnings with specs on modern Rubies
- [x] Remove unnecessary/outddated cruft from the Maven build process
- [x] Update dependencies where possible
- [x] Fix other test failures
  - Moves Rack back to `2.2.9` as the last known validated-in-real-world version (?) which passes the tests.
  - Drops rest for Rubygems-based "magic comment" install which seems difficult to get to work now that rubygems is automatically loaded during init on modern Rubies
  
**Changes in non-test functionality**
- [x] Change jruby-core dependency back to `provided` scope, to allow for users why may be using `jruby-complete`. BYO JRuby, basically.
- [x] Remove Log4JLogger (1.x) which is EOL, insecure and should not be used  
- [x] Drops loadpath hacks/workarounds for misbehaving J2EE servers in `booter.rb` which causes problems with getting tests to pass and seems extremely legacy now.


chadlwilson:
Rack 3.1.3
- 570 examples, 215 failures

Rack 3.0.11 (current master)
- 570 examples, 120 failures

Rack 2.2.9, 1.6.13, 1.3.10
- 570 examples, 116 failures

Will temporarily go back to Rack 2.2.9 and work forward I think, as also note there is some logic to try and dynamically install and test against specific (old) Rack versions.

chadlwilson:
OK down now to `570 examples, 14 failures` with Rack `2.2.9`

```
[INFO] Failed examples:
[INFO]
[INFO] rspec ./src/spec/ruby/jruby/rack/booter_spec.rb:244 # JRuby::Rack::Booter adjusts load path when runtime.jruby_home == /tmp
[INFO] rspec ./src/spec/ruby/rack/application_spec.rb:169 # Java::OrgJrubyRack::DefaultRackApplicationFactory should init and create application object without a rackup script
[INFO] rspec ./src/spec/ruby/rack/application_spec.rb:313 # Java::OrgJrubyRack::DefaultRackApplicationFactory initialized newRuntime loads specified version of rack
[INFO] rspec ./src/spec/ruby/rack/application_spec.rb:428 # Java::OrgJrubyRack::DefaultRackApplicationFactory initialized newRuntime allows to keep RUBYOPT with a clear environment
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:1:6]' # JRuby::Rack::Input rewindable should read into a provided buffer
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:1:7]' # JRuby::Rack::Input rewindable should read a specified amount into a provided buffer
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:1:8]' # JRuby::Rack::Input rewindable should replace contents of buffer
[INFO] rspec ./src/spec/ruby/rack/input_spec.rb:109 # JRuby::Rack::Input rewindable reads an image
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:2:6]' # JRuby::Rack::Input rewindable (with buffer size) should read into a provided buffer
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:2:7]' # JRuby::Rack::Input rewindable (with buffer size) should read a specified amount into a provided buffer
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:2:8]' # JRuby::Rack::Input rewindable (with buffer size) should replace contents of buffer
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:4:6]' # JRuby::Rack::Input non-rewindable should read into a provided buffer
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:4:7]' # JRuby::Rack::Input non-rewindable should read a specified amount into a provided buffer
[INFO] rspec './src/spec/ruby/rack/input_spec.rb[1:4:8]' # JRuby::Rack::Input non-rewindable should replace contents of buffer
```

@headius would you mind taking a look at the `JRuby::Rack::Input` errors? I presume related to #256?

The additional 4 failures with Rack 3.0 are the **rack-compat** tests:

```
[INFO] rspec './src/spec/ruby/rack/handler/servlet_spec.rb[1:1:1:24]' # Rack::Handler::Servlet (default) env behaves like env parses strange request parameters (Rack-compat)
[INFO] rspec './src/spec/ruby/rack/handler/servlet_spec.rb[1:1:1:25]' # Rack::Handler::Servlet (default) env behaves like env parses nestedx request parameters (Rack-compat)
[INFO] rspec './src/spec/ruby/rack/handler/servlet_spec.rb[1:2:1:24]' # Rack::Handler::Servlet lazy env behaves like env parses strange request parameters (Rack-compat)
[INFO] rspec './src/spec/ruby/rack/handler/servlet_spec.rb[1:2:1:25]' # Rack::Handler::Servlet lazy env behaves like env parses nestedx request parameters (Rack-compat)
```

headius:
I'll have a look at the Input failures soon.

headius:
I suppose that's what I get for not getting tests running before making changes. Here's the patch for JRuby::Rack::Input:

```diff
diff --git a/src/main/java/org/jruby/rack/ext/Input.java b/src/main/java/org/jruby/rack/ext/Input.java
index e680e79..2cdd8e3 100644
--- a/src/main/java/org/jruby/rack/ext/Input.java
+++ b/src/main/java/org/jruby/rack/ext/Input.java
@@ -172,7 +172,7 @@ public class Input extends RubyObject {
                 if ( buffer != null ) {
                     buffer.clear();
                     try {
-                        int unused = (int) CONCAT_WITH_CODERANGE.invokeExact(new ByteList(bytes, false), StringSupport.CR_UNKNOWN);
+                        int unused = (int) CONCAT_WITH_CODERANGE.invokeExact(buffer, new ByteList(bytes, false), StringSupport.CR_UNKNOWN);
                     } catch (Throwable t) {
                         Helpers.throwException(t);
                     }
```

headius:
@jlahtinen The patch above should work properly... I botched it before 🤦.

chadlwilson:
Thanks @headius - done now. The remaining failing four are a bit of a mystery to me right now.

In the original failure messages, a couple below **actually** report useless (`assigning non-exception to $!`) errors that obscure the *real* errors (at root, they are all `org.jruby.exceptions.LoadError: (Missing SpecError) Gem::MissingSpecError`s). Some Ruby/JRuby/rspec/jruby-rack problem I guess?

Locally I temporarily did a `try { } catch (Throwable e) { throw new RuntimeException(e); }` around the original code temporarily to workaround whatever is causing the exceptions to be lost and figure out the root.

https://github.com/jruby/jruby-rack/blob/e49226340b2b8c9796beea7cfe31830229ec582d/src/main/java/org/jruby/rack/DefaultRackApplicationFactory.java#L430

```
[INFO] Failures:
[INFO]
[INFO]   1) JRuby::Rack::Booter adjusts load path when runtime.jruby_home == /tmp
[INFO]      Failure/Error: booter.boot!
[INFO]
[INFO]      LoadError:
[INFO]        no such file to load -- jruby
[INFO]        Did you mean?  jruby/ext
[INFO]      # org/jruby/RubyKernel.java:1184:in `require'
[INFO]      # uri:classloader:/jruby/rack/booter.rb:177:in `adjust_load_path'
[INFO]      # uri:classloader:/jruby/rack/booter.rb:93:in `boot!'
[INFO]      # ./src/spec/ruby/jruby/rack/booter_spec.rb:258:in `block in <main>'
[INFO]      # org/jruby/RubyBasicObject.java:2620:in `instance_exec'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:263:in `block in run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511:in `block in with_around_and_singleton_context_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468:in `block in with_around_example_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:486:in `block in run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:624:in `run_around_example_hooks_for'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:486:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468:in `with_around_example_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511:in `with_around_and_singleton_context_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:259:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:646:in `block in run_examples'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:651:in `run_examples'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:607:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121:in `block in run_specs'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121:in `block in run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/configuration.rb:2091:in `with_suite_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:116:in `block in run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/reporter.rb:74:in `report'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:115:in `run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:89:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:71:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:45:in `invoke'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/exe/rspec:4:in `<main>'
[INFO]      # org/jruby/RubyKernel.java:1220:in `load'
[INFO]
[INFO]   2) Java::OrgJrubyRack::DefaultRackApplicationFactory should init and create application object without a rackup script
[INFO]      Failure/Error: fail "missing run or map statement" unless app
[INFO]
[INFO]      RuntimeError:
[INFO]        missing run or map statement
[INFO]      # ./target/rubygems/gems/rack-2.2.9/lib/rack/builder.rb:233:in `to_app'
[INFO]      # uri:classloader://config.ru:1:in `<main>'
[INFO]      # uri:classloader://config.ru:1:in `block in <main>'
[INFO]      # org/jruby/RubyBasicObject.java:2620:in `instance_exec'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:263:in `block in run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511:in `block in with_around_and_singleton_context_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468:in `block in with_around_example_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:486:in `block in run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:624:in `run_around_example_hooks_for'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:486:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468:in `with_around_example_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511:in `with_around_and_singleton_context_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:259:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:646:in `block in run_examples'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:651:in `run_examples'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:607:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121:in `block in run_specs'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121:in `block in run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/configuration.rb:2091:in `with_suite_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:116:in `block in run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/reporter.rb:74:in `report'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:115:in `run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:89:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:71:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:45:in `invoke'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/exe/rspec:4:in `<main>'
[INFO]      # org/jruby/RubyKernel.java:1220:in `load'
[INFO]
[INFO]   3) Java::OrgJrubyRack::DefaultRackApplicationFactory initialized newRuntime loads specified version of rack
[INFO]      Failure/Error: Unable to find org.jruby.rack.DefaultRackApplicationFactory.checkAndSetRackVersion(org/jruby/rack/DefaultRackApplicationFactory.java to read failed line
[INFO]
[INFO]      Java::JavaLang::RuntimeException:
[INFO]        org.jruby.exceptions.LoadError: (MissingSpecError) Gem::MissingSpecError
[INFO]      # org.jruby.rack.DefaultRackApplicationFactory.checkAndSetRackVersion(org/jruby/rack/DefaultRackApplicationFactory.java:434)
[INFO]      # java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:498)
[INFO]      # org.jruby.javasupport.JavaMethod.invokeDirectWithExceptionHandling(org/jruby/javasupport/JavaMethod.java:315)
[INFO]      # org.jruby.javasupport.JavaMethod.invokeDirect(org/jruby/javasupport/JavaMethod.java:176)
[INFO]      # RUBY.<main>(/Users/chad/Projects/community/jruby-rack/src/spec/ruby/rack/application_spec.rb:323)
[INFO]      # org.jruby.RubyBasicObject.yieldUnder(org/jruby/RubyBasicObject.java:1736)
[INFO]      # org.jruby.RubyBasicObject.instance_exec(org/jruby/RubyBasicObject.java:2620)
[INFO]      # org.jruby.RubyBasicObject$INVOKER$i$0$0$instance_exec.call(org/jruby/RubyBasicObject$INVOKER$i$0$0$instance_exec.gen)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.invokeOther4:instance_exec(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:263)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:263)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.with_around_and_singleton_context_hooks(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.with_around_example_hooks(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.hooks.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:486)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.hooks.run_around_example_hooks_for(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:624)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.hooks.invokeOther10:run_around_example_hooks_for(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:486)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.hooks.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/hooks.rb:486)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.invokeOther2:run(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.with_around_example_hooks(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.invokeOther4:with_around_example_hooks(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.with_around_and_singleton_context_hooks(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.invokeOther42:with_around_and_singleton_context_hooks(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:259)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:259)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.invokeOther6:run(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:646)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.run_examples(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:646)
[INFO]      # org.jruby.RubyArray.collectArray(org/jruby/RubyArray.java:2770)
[INFO]      # org.jruby.RubyArray.map(org/jruby/RubyArray.java:2803)
[INFO]      # org.jruby.RubyArray$INVOKER$i$0$0$map.call(org/jruby/RubyArray$INVOKER$i$0$0$map.gen)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.invokeOther14:map(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:651)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.run_examples(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:651)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.invokeOther24:run_examples(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:607)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:607)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.invokeOther0:run(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # org.jruby.RubyArray.collectArray(org/jruby/RubyArray.java:2770)
[INFO]      # org.jruby.RubyArray.map(org/jruby/RubyArray.java:2803)
[INFO]      # org.jruby.RubyArray$INVOKER$i$0$0$map.call(org/jruby/RubyArray$INVOKER$i$0$0$map.gen)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.invokeOther28:map(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.invokeOther0:run(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # org.jruby.RubyArray.collectArray(org/jruby/RubyArray.java:2770)
[INFO]      # org.jruby.RubyArray.map(org/jruby/RubyArray.java:2803)
[INFO]      # org.jruby.RubyArray$INVOKER$i$0$0$map.call(org/jruby/RubyArray$INVOKER$i$0$0$map.gen)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.invokeOther28:map(Users/chad/Projects/community/jruby_minus_rack/target/rubygems/gems/rspec_minus_core_minus_3_dot_13_dot_0/lib/rspec/core//Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # Users.chad.Projects.community.jruby_minus_rack.target.rubygems.gems.rspec_minus_core_minus_3_dot_13_dot_0.lib.rspec.core.example_group.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608)
[INFO]      # RUBY.run_specs(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121)
[INFO]      # org.jruby.RubyArray.collectArray(org/jruby/RubyArray.java:2770)
[INFO]      # org.jruby.RubyArray.map(org/jruby/RubyArray.java:2803)
[INFO]      # org.jruby.RubyArray$INVOKER$i$0$0$map.call(org/jruby/RubyArray$INVOKER$i$0$0$map.gen)
[INFO]      # RUBY.run_specs(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121)
[INFO]      # RUBY.with_suite_hooks(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/configuration.rb:2091)
[INFO]      # RUBY.run_specs(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:116)
[INFO]      # RUBY.report(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/reporter.rb:74)
[INFO]      # RUBY.run_specs(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:115)
[INFO]      # RUBY.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:89)
[INFO]      # RUBY.run(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:71)
[INFO]      # RUBY.invoke(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:45)
[INFO]      # RUBY.<main>(/Users/chad/Projects/community/jruby-rack/target/rubygems/gems/rspec-core-3.13.0/exe/rspec:4)
[INFO]      # org.jruby.Ruby.runInterpreter(org/jruby/Ruby.java:1258)
[INFO]      # org.jruby.Ruby.loadFile(org/jruby/Ruby.java:2977)
[INFO]      # org.jruby.RubyKernel.loadCommon(org/jruby/RubyKernel.java:1250)
[INFO]      # org.jruby.RubyKernel.load(org/jruby/RubyKernel.java:1220)
[INFO]      # org.jruby.RubyKernel$INVOKER$s$load.call(org/jruby/RubyKernel$INVOKER$s$load.gen)
[INFO]      # java.lang.invoke.MethodHandle.invokeWithArguments(java/lang/invoke/MethodHandle.java:627)
[INFO]      # org.jruby.Ruby.runScript(org/jruby/Ruby.java:1245)
[INFO]      # org.jruby.Ruby.runNormally(org/jruby/Ruby.java:1157)
[INFO]      # org.jruby.Ruby.runFromMain(org/jruby/Ruby.java:983)
[INFO]      # org.jruby.Main.doRunFromMain(org/jruby/Main.java:398)
[INFO]      # org.jruby.Main.internalRun(org/jruby/Main.java:282)
[INFO]      # org.jruby.Main.run(org/jruby/Main.java:227)
[INFO]      # org.jruby.Main.main(org/jruby/Main.java:199)
[INFO]      # ------------------
[INFO]      # --- Caused by: ---
[INFO]      # Java::OrgJrubyExceptions::LoadError:
[INFO]      #   (MissingSpecError) Gem::MissingSpecError
[INFO]      #   RUBY.to_specs(uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/dependency.rb:311)
[INFO]
[INFO]   4) Java::OrgJrubyRack::DefaultRackApplicationFactory initialized newRuntime allows to keep RUBYOPT with a clear environment
[INFO]      Failure/Error: hooks.run(:around, :example, self) { yield }
[INFO]
[INFO]      TypeError:
[INFO]        assigning non-exception to $!
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:468:in `with_around_example_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:511:in `with_around_and_singleton_context_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example.rb:259:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:646:in `block in run_examples'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:651:in `run_examples'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:607:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608:in `block in run'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608:in `block in run'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/example_group.rb:608:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121:in `block in run_specs'
[INFO]      # org/jruby/RubyArray.java:2803:in `map'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:121:in `block in run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/configuration.rb:2091:in `with_suite_hooks'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:116:in `block in run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/reporter.rb:74:in `report'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:115:in `run_specs'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:89:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:71:in `run'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/lib/rspec/core/runner.rb:45:in `invoke'
[INFO]      # ./target/rubygems/gems/rspec-core-3.13.0/exe/rspec:4:in `<main>'
[INFO]      # org/jruby/RubyKernel.java:1220:in `load'
[INFO]
[INFO] 2 deprecation warnings total
[INFO]
[INFO] Finished in 27.71 seconds (files took 1.33 seconds to load)
[INFO] 570 examples, 4 failures
```

chadlwilson:
Hmm, there's a few suspicious bits around the place like the below, so perhaps these specs requiring a new rack version to be dynamically installed or use of JRuby home in tmpdir  weren't working for a long time - not sure how long since TravisCI was passing :D

https://github.com/jruby/jruby-rack/blob/5c060aeef4afa5b3e6eee13bd3429675cf17c16f/.travis.yml#L38-L39

https://github.com/jruby/jruby-rack/blob/d3ee8f04302856db83c36f8825355584be23c4f4/src/main/ruby/jruby/rack/booter.rb#L176-L192

enebo:
@chadlwilson that first loaderror is pretty odd.  If I try and load it I can see it:

```text
java -jar jruby-complete-9.4.7.0.jar -r jruby -e 'p $LOADED_FEATURES'
["java.rb", "jruby/util.rb", "enumerator.rb", "rational.rb", "complex.rb", "thread.rb", "fiber.rb", "ruby2_keywords.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/unicode_normalize.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rbconfig.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/compatibility.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/defaults.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/deprecate.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/errors.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/unknown_command_spell_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/exceptions.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/basic_specification.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/stub_specification.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/platform.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/util/list.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/version.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/requirement.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/specification.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/jars/post_install_hook.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/jar_install_post_install_hook.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/defaults/jruby.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/defaults/operating_system.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/util.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/dependency.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_gem.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/monitor.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_warn.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/path_support.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/version.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/core_ext/name_error.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/levenshtein.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/jaro_winkler.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/name_error_checkers/class_name_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/name_error_checkers/variable_name_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/name_error_checkers.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/method_name_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/key_error_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/null_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/tree_spell_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/require_path_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/spell_checkers/pattern_key_name_checker.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean/formatter.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/did_you_mean.rb", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/jruby.rb"]
```

If I had to guess perhaps something is changing `$LOAD_PATH`?

```text
java -jar jruby-complete-9.4.7.0.jar -r jruby -e 'p $LOAD_PATH'
["uri:classloader:/META-INF/jruby.home/lib/ruby/3.1/site_ruby", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib"]
```

enebo:
Unrelated to `jruby` not loading but all that code after that point assumes we are either Ruby 1.8 or Ruby 1.9 implementation but we have be a 2+ impl for like a decade so I think that code is going to be broken.  In my previous comment if you see site_ruby it is is in `ruby/${ruby_major}/site_ruby` and not shared.  I think there will be an opportunity to delete a lot of that method.

chadlwilson:
Thanks for poking your nose in @enebo !

Yeah, I realised that code was ancient/outdated and tried removing the various 1.7 and 1.8 fallback stuff, but wasn't sure what it should be on modern JRubies or Rubies or even entirely the intent of the test or code :D The test itself is changing the loadpath:

https://github.com/jruby/jruby-rack/blob/942597332def78467d415ea6ed9b95b35db3b222/src/spec/ruby/jruby/rack/booter_spec.rb#L241-L258

Given my personal usage of jruby-rack works fine with `9.4.7.0` I assume this booter is not used in all cases which is why I havent seen such problems in the real world.

I don't really understand how the test is supposed to work, given it hacks the `LOAD_PATH` to tmp before `require 'jruby'`. I suppose that's perhaps why there is an earlier explicit `require 'jruby'` in the spec?

Original commit/hackery here: https://github.com/jruby/jruby-rack/commit/3c523c5cdc30a99fd9da634ac7c1b9cdc6f8ef9d

enebo:
@chadlwilson I am also guessing the booter does not remove the include directory that has jruby.rb in it either.  This spec does though.  Random thoughts:
 
 1. no point in making require 'jruby' be in the method.  It is so likely to get called by something else in this particular library it may as well just be at top of file.  
 2. Remove all the version checking which only leaves the jruby_home call (which is too bad because that is what will still require `jruby` as a require.
 3. What is the appropriate paths to index on?  In this case:

```text
uri:classloader:/META-INF/jruby.home/lib/ruby/3.1/site_ruby", "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib"
```

it is ``["3.1/site_ruby", "stdlib"]`.  This may need rbconfig to get version so this is more future proof.

 4. Does it make sense to even have this method any more?  @kares might know if newer web containers are well-behaved or perhaps whether we can just back off and see if a websphere user exists and needs it fixed again.  My opinion is this code is ancient and it is a good opportunity to try and delete some stuff.  This method has not worked for anyone using this since we supported Ruby 1.9 (I think this because we changed lib/ruby layout during Ruby 2).

Anyways some thoughts to try and help.  This work is incredible and we need to get jruby-rack back into shape.

chadlwilson:
Thanks, yeah. I tried changing the paths after your message but it still seemed to be failing in the same way.

Actually after reading your message it reminded me that I am trying to do this on `master` which is actually missing years of fixes to `1.1` right now currently pending in #253 

My thinking was to try and get the tests passing to give more confidence in merging that PR, but given 1.1 is probably more stable than 1.2/master (right now) and may indeed have test fixes, I might have that around the wrong way and should at least look at that branch.

It's a separate topic to this specific case (where deleting the old code seems most advisable) but perhaps worth keeping in mind for some of the remaining failures.

chadlwilson:
The last remaining test failing is https://github.com/jruby/jruby-rack/blob/cb356674de495300544a32d47bbb4bf72cbbfeee/src/spec/ruby/rack/application_spec.rb#L313-L328

I am not a Ruby expert, but I wonder if the underlying functionality is no longer required now.

This functionality appears to allow jruby-rack to dynamically install/load/require a specific Rack version from `config.ru` using Rubygems alone, using a magic comment - **without** `Gemfile`/bundler being available. This was introduced in https://github.com/jruby/jruby-rack/commit/32b531ff3b5bbe6b13672cac3d6a432e52931be8

There is also a bundled version of rack which I don't fully understand interactions of. Not quite sure how to proceed on this one, but suffice it to say that after this logic is executed in test setup without error, rack can still not be found

https://github.com/jruby/jruby-rack/blob/34f390c33106427f628640872a92a9e105cb1adf/src/spec/ruby/spec_helper.rb#L96-L100

...leading to...

```
[INFO]   1) Java::OrgJrubyRack::DefaultRackApplicationFactory initialized newRuntime loads specified version of rack
[INFO]      Failure/Error: Unable to find org.jruby.rack.DefaultRackApplicationFactory.checkAndSetRackVersion(org/jruby/rack/DefaultRackApplicationFactory.java to read failed line
[INFO]
[INFO]      Java::JavaLang::RuntimeException:
[INFO]        org.jruby.exceptions.LoadError: (MissingSpecError) Gem::MissingSpecError
```

when running

https://github.com/jruby/jruby-rack/blob/e49226340b2b8c9796beea7cfe31830229ec582d/src/main/java/org/jruby/rack/DefaultRackApplicationFactory.java#L430

enebo:
@chadlwilson sorry I missed your last comments.   I am not sure if this is still valid or not and I do not see any obvious rack documentation on this (although to be honest I only looked a for a few minutes).  It is definitely possible to run JRuby without bundler but has rack just decided people need bundler?  Does the magic comments still work?  Maybe @kares knows.

A possible way to test this would be to make a simple config.ru and trying it with C Ruby and see if it works with the magic comment.  The fact this is just embedded code in Java does not change that this could just be invoked from Ruby.  If it works then we probably need to change how we require the gem (like something is missing or gem api changed over time and we are not calling something).  If it doesn't then it can be nuked from orbit.

chadlwilson:
No worries @enebo 

I think the magic comment is a jruby-rack invention, perhaps specific to a Jruby context due jruby-rack bundling an opinionated rack version: https://github.com/jruby/jruby-rack/blob/master/History.md#1112-281112

> with a plain Rack app when config.ru is evaluated it already assumes Rack to be loaded hovewer it's common practice to `bundle exec rackup` such scripts while Bundler loads a version of Rack that might be different than JRuby-Rack's bundled Rack version. 
>
> thus for now we support a magic comment :
> `# rack.version = ~>1.3.6`
> ...
> with "special" support for Bundler as well :
> `# encoding: UTF-8`
> `# rack.version = bundler`

I'm a little bit lost here as I don't know rackup, but it seems that the intent here is to allow an arbitrary rack version with an external bundler rackup, but then use the magic comment to run whatever the jruby-rack bundled version is.

So as long as rackup is still useful, and jruby-rack still bundles a rack version, it seems conceptually possible that the functionality is needed, but I can't for the life of me figure out how to empirically test the various cases and work backwards to how the test should behave 😅 (definitely a skill issue, rack is not something I've put much time into understanding)

enebo:
@chadlwilson I decided to look a little bit.  In trying to just run rake instead of full maven run I had to tweak:

```diff
diff --git a/Rakefile b/Rakefile
index 599ce7b..19f50a2 100644
--- a/Rakefile
+++ b/Rakefile
@@ -110,7 +110,8 @@ task :speconly => [ :resources, :test_resources ] do
     spec = ENV['SPEC'] || File.join(Dir.getwd, "src/spec/ruby/**/*_spec.rb")
     opts = opts.split(' ').push *FileList[spec].to_a
 
-    args = ENV['RUBY'].split(' ') + %w{-Isrc/spec/ruby -rbundler/setup -S rspec}
+    ruby = ENV['RUBY'] || 'jruby'
+    args = ruby.split(' ') + %w{-Isrc/spec/ruby -rbundler/setup -S rspec}
     sh *args, *opts
   end
 end
```

Then rake test runs but I get 3 errors all of which is failing to find servlet class so servlet_30? fails to work right.  I guess this may be a stretch goal but it would be nice to just run rake for local dev.  Maven does repro locally for me so something with maven it properly loading the servlet API before running rake.

enebo:
Ah I see ... this will generate a full Java CMD and not use jruby itself.  This also explains why servlets classes are not found since it builds up a classpath while in maven.  Ignore my previous comment.  It would be nice just to type `rake` but this was not designed around being able to do that

chadlwilson:
> Ah I see ... this will generate a full Java CMD and not use jruby itself. This also explains why servlets classes are not found since it builds up a classpath while in maven. Ignore my previous comment. It would be nice just to type `rake` but this was not designed around being able to do that

Yeah, I agree. The feedback loop is pretty painful right now. I was just going to path of least resistance since things were not in a working state, and I had no access to prior builds to even know when the tests last passed, against which versions of things and on which branches :-) This branch also does not have a whole lot of fixes from https://github.com/jruby/jruby-rack/tree/1.1-stable just yet, so it's possible that will change things again.

It's also possible that a change somewhere on that branch not merged to master did something to address the tests here, or even allow them to work with `rake`, so I should probably pause and take a look at that later.

chadlwilson:
> 
> ```diff
> diff --git a/Rakefile b/Rakefile
> index 599ce7b..19f50a2 100644
> --- a/Rakefile
> +++ b/Rakefile
> @@ -110,7 +110,8 @@ task :speconly => [ :resources, :test_resources ] do
>      spec = ENV['SPEC'] || File.join(Dir.getwd, "src/spec/ruby/**/*_spec.rb")
>      opts = opts.split(' ').push *FileList[spec].to_a
>  
> -    args = ENV['RUBY'].split(' ') + %w{-Isrc/spec/ruby -rbundler/setup -S rspec}
> +    ruby = ENV['RUBY'] || 'jruby'
> +    args = ruby.split(' ') + %w{-Isrc/spec/ruby -rbundler/setup -S rspec}
>      sh *args, *opts
>    end
>  end
> ```
> 

Thx, I've merged this - that's better.

> Then rake test runs but I get 3 errors all of which is failing to find servlet class so servlet_30? fails to work right. I guess this may be a stretch goal but it would be nice to just run rake for local dev. Maven does repro locally for me so something with maven it properly loading the servlet API before running rake.

Interesting, this provides a hint I guess to why the other test is failing, as it must be to do with the wrapping environment `mvn` -> `rake` -> `mvn` or something I changed here. I'll take another look tomorrow perhaps with this in mind...

enebo:
The `ENV['RUBY']` when executed from maven is: 

```text
/home/enebo/Applications/jdks/jdk8u402-b06/jre/bin/java -cp /home/enebo/.m2/repository/org/jruby/jruby-complete/9.4.8.0-SNAPSHOT/jruby-complete-9.4.8.0-SNAPSHOT.jar:/home/enebo/work/gems/jruby-rack/target/test-classes:/home/enebo/work/gems/jruby-rack/target/classes:/home/enebo/.m2/repository/org/jruby/jruby-core/9.4.8.0-SNAPSHOT/jruby-core-9.4.8.0-SNAPSHOT.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-servlet_3.0_spec/1.0/geronimo-servlet_3.0_spec-1.0.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-jsp_2.2_spec/1.2/geronimo-jsp_2.2_spec-1.2.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-el_2.2_spec/1.0.4/geronimo-el_2.2_spec-1.0.4.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-jms_1.1_spec/1.1.1/geronimo-jms_1.1_spec-1.1.1.jar:/home/enebo/.m2/repository/commons-logging/commons-logging-api/1.1/commons-logging-api-1.1.jar:/home/enebo/.m2/repository/org/slf4j/slf4j-api/1.7.2/slf4j-api-1.7.2.jar:/home/enebo/.m2/repository/log4j/log4j/1.2.17/log4j-1.2.17.jar:/home/enebo/.m2/repository/org/springframework/spring-core/3.2.10.RELEASE/spring-core-3.2.10.RELEASE.jar:/home/enebo/.m2/repository/commons-logging/commons-logging/1.1.3/commons-logging-1.1.3.jar:/home/enebo/.m2/repository/javax/activation/activation/1.1/activation-1.1.jar:/home/enebo/.m2/repository/org/jruby/jruby-complete/9.4.8.0-SNAPSHOT/jruby-complete-9.4.8.0-SNAPSHOT.jar org.jruby.Main
```

So it knows how to acculmulate all those jar files into a java command.  My patch makes things work until you need one of these jars (since just calling jruby by itself is not adding them).  It is fascinating that only three things fail though.  If whatever jar which includes `javax.servlet.AsyncContext` then MAYBE things will work:

```ruby
  def servlet_30?
    @@servlet_30 = !! ( Java::JavaClass.for_name('javax.servlet.AsyncContext') rescue nil )
  end
```

So the specs could load the jar(s) via require...but then we would have this odd data replication since we specify deps in pom.xml and we would also have requires elsewhere.  I might have opened a can of worms :)

chadlwilson:
> The `ENV['RUBY']` when executed from maven is:
> 
> ```
> /home/enebo/Applications/jdks/jdk8u402-b06/jre/bin/java -cp /home/enebo/.m2/repository/org/jruby/jruby-complete/9.4.8.0-SNAPSHOT/jruby-complete-9.4.8.0-SNAPSHOT.jar:/home/enebo/work/gems/jruby-rack/target/test-classes:/home/enebo/work/gems/jruby-rack/target/classes:/home/enebo/.m2/repository/org/jruby/jruby-core/9.4.8.0-SNAPSHOT/jruby-core-9.4.8.0-SNAPSHOT.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-servlet_3.0_spec/1.0/geronimo-servlet_3.0_spec-1.0.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-jsp_2.2_spec/1.2/geronimo-jsp_2.2_spec-1.2.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-el_2.2_spec/1.0.4/geronimo-el_2.2_spec-1.0.4.jar:/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-jms_1.1_spec/1.1.1/geronimo-jms_1.1_spec-1.1.1.jar:/home/enebo/.m2/repository/commons-logging/commons-logging-api/1.1/commons-logging-api-1.1.jar:/home/enebo/.m2/repository/org/slf4j/slf4j-api/1.7.2/slf4j-api-1.7.2.jar:/home/enebo/.m2/repository/log4j/log4j/1.2.17/log4j-1.2.17.jar:/home/enebo/.m2/repository/org/springframework/spring-core/3.2.10.RELEASE/spring-core-3.2.10.RELEASE.jar:/home/enebo/.m2/repository/commons-logging/commons-logging/1.1.3/commons-logging-1.1.3.jar:/home/enebo/.m2/repository/javax/activation/activation/1.1/activation-1.1.jar:/home/enebo/.m2/repository/org/jruby/jruby-complete/9.4.8.0-SNAPSHOT/jruby-complete-9.4.8.0-SNAPSHOT.jar org.jruby.Main
> ```
> 
> So it knows how to acculmulate all those jar files into a java command. My patch makes things work until you need one of these jars (since just calling jruby by itself is not adding them). It is fascinating that only three things fail though. If whatever jar which includes `javax.servlet.AsyncContext` then MAYBE things will work:

Yeah, on current deps it will be `/home/enebo/.m2/repository/org/apache/geronimo/specs/geronimo-servlet_3.0_spec/1.0/geronimo-servlet_3.0_spec-1.0.jar` but given all the other `provided` dependencies, it's a bit surprising there aren't more things actually being used. I think perhaps for the examples or appraisals or whatnot which I haven't looked at.

> So the specs could load the jar(s) via require...but then we would have this odd data replication since we specify deps in pom.xml and we would also have requires elsewhere. I might have opened a can of worms :)

Since rake already calls some `mvn` tasks to set things up, probably just as easy to get it to to dump a classpath for the `test` lifecycle somehow and then use that in the `Rakefile`. But ya, probably a can of worms :-)

chadlwilson:
The problem with the last failing test when run via `mvn test` (compared to `bundle exec rake test` where it is OK) seems to be that gems are being loaded from a different place than they have been  earlier installed to by the test code.

When running via Maven, the test installs the "override" rack version to a place like `GEM_HOME=/Users/chad/Projects/community/jruby-rack/target/rubygems` inside the maven `build.directory`.

Before requiring `rubygems` as part of initializating rack, the code then tries to override `GEM_HOME` within the newly created "under-test" jruby runtime, however it is unsuccessful in altering the `Gem.dir` which remains `uri:classloader:/META-INF/jruby.home/lib/ruby/gems/shared` inside the jruby-core jar.

https://github.com/jruby/jruby-rack/blob/39cdba571bb964599ce8209ff1a8c0c162e8c0d0/src/spec/ruby/rack/application_spec.rb#L320-L324

These first two lines dont seem to have the intended effect of changing where the runtime will load Gems from to match what `rake` is running with. So it tries to load from a different location than the Gem was earlier installed, and thus you get `MissingSpecError`.

By contrast, when running tests directly via `rake` the `GEM_HOME`/`Gem.dir` ends up ones external jruby installation (e.g `/Users/chad/.local/share/mise/installs/ruby/jruby-9.4.7.0/lib/ruby/gems/shared`) and since it matches defaults it works OK. For some reason, this problem doesn't seem to affect something similar done in a Bundler test, but haven't dug into why that one works.

@enebo or @headius - once against excusing my JRuby ignorance, does this ring a bell to you, as to why the below would be ineffective in actually changing `Gem.dir` and affecting where RubyGems try to load from? Possibly it has already been loaded or this logic only worked on earlier Ruby/JRuby versions.

``` ruby
@runtime.evalScriptlet "ENV['GEM_HOME'] = #{ENV['GEM_HOME'].inspect}"
```

enebo:
@chadlwilson I realize after reading this that rubygems is run as part of initialization and this used to not be the case (we changed to this whenever C Ruby made the change).    This spec starts a new Ruby instances which loads rubygems then it sets GEM_HOME after gems have already loaded. as part of the instance coming up.  This explains why it is not seeing the gems.

We can probably make this work by setting java system property `jruby.cli.rubygems.enable` to false before creating the instance but I am questioning the value of this test in a world where rubygems is just loaded by default.  All this test is really doing is verifying Ruby gems can change its home location.  I think it would be safe to just remove this test and if there is some future issue about people not being able to customize GEM_HOME we maybe start again.

enebo:
@chadlwilson to be more explicit I mean this line will load rubygems: https://github.com/jruby/jruby-rack/blob/39cdba571bb964599ce8209ff1a8c0c162e8c0d0/src/spec/ruby/rack/application_spec.rb#L319

chadlwilson:
Right, that's what I thought might be happening.

The test is really trying to check you can load a custom rack version using the magic comments and needs to change the GEM_HOME to do so between tests. Alternatively it could change it for all tests prior to starting and have the same effect rather than trying to swap it in between tests.

If there is no way to change the Gem.dir after rubygems is loaded what would be the right way to have the GEM_HOME or additional GEM_PATHs set prior to the runtime creation when doing so programmatically like in these tests so it's picked up by the initialization?

enebo:
@chadlwilson oh yeah I forgot!  You can try something like this before the newRuntime call (but you should remove it after the test as well):

```ruby
java.lang.System.properties.put("jruby.cli.rubygems.enable", "false")
```

This should prevent it from loading rubygems by default...which is unclear how gems ends up being loaded by the test but it must do it somewhere.

chadlwilson:
Sadly I cannot find a place to set this (or any related things!) that seems to work. The runtimes are reused between tests, but since it is all happening within one JVM instance launched by `mvn test --> rake` I actually dont quite understand why the `GEM_HOME` originally set for the wrapping process doesn't auto-magically propagate through to how JRuby is configured when it is required and/or runtimes created.

I also tried appending to `Gem.path` manually via `Gem.path.unshift`, which shows as updated but not effective in allowing the Gem to load.

Due to my own inexperience, I am leaning towards just deleting this test and leaving it untested, relying solely on the bundler test and assuming that any sane individual trying to do this magic comment stuff is doing so via Bundler/Gemfile rather than relying on a naked Gem install.

enebo:
@chadlwilson I think excluding and working towards the longer goal of getting jruby-rack back up to date and in shape.  We can always figure this out later (pend the spec).

chadlwilson:
OK, I have removed this one test, which gets things passing now. Ready for review whenever you folks are able.

Next I'll try and tackle getting all the `1.1` fixes/changes merged back onto master with the tests in a known state.

enebo:
Everything looked so straightforward I just merged this without other reviews (which I am not normally a jruby-rack person but these were all sensible changes).  @chadlwilson Thanks.  You are well on your way to untangling this thing.  We look forward to the next step.

PR #266, opened 2024-10-24
feat: Jakarta EE 9+ migration

### LINKED ISSUE #234: Support EE9 / jakarta servlet specs
I'm maintaining a project that was developed using JRuby and rack - deployed onto tomcat.

With the latest break-the-world change:
https://www.i-programmer.info/news/80-java/14216-jakarta-ee-9-released.html
This project is now useless for Tomcat 10, and beyond.

Is this project dead / will never support the servlet specification rename?

Or is there a better / newer way to take JRuby code and build a war that will run in Tomcat 10?

bhagyashriSawkar:
Updated a few files with changes required for upgrade to Jakarta EE 9+. The jar is getting generated but test cases are not getting executed, will work on that.

Please let me know if there is something else that also needs to be checked. Also I did override a few required abstract method with empty/default value, would appreciate if you take a look at those.

Thanks.


chadlwilson:
This would be a breaking change probably requiring a `1.3` or `2.0` release.

Right now `1.2.x` off master is not very stable (needs #262 at the very least to get fixes that were already there on `1.1.x`), and still lacks validated testing with modern Rails versions for which we probably need help with to get stable.

If you know Rails and jruby-rack better than me (especially the session stuff) I could do with some help trying to get the tests working with modern Rails and JRuby versions. I'll probably create a separate draft PR for that shortly.

chadlwilson:
The branch/PR at #262 does get the tests/specs all running/passing reliably however :-)

kares:
:+1: looks like a good start but the PR feels incomplete.

my idea about supporting jakarta/javax.servlet at the same time would have been converging towards moving all .rb bits that script the javax.servlet API into .java and than potentially use some kind of a weaving approach or the ["migration"](https://tomcat.apache.org/download-migration.cgi) tool than Tomcat has to release 2 artifacts from the same build... :shrug:

headius:
Exciting! This will obviously need to be rebased on master due to recent updates there, but I'm glad you've attempted this work.

> This would be a breaking change probably requiring a 1.3 or 2.0 release.

> use some kind of a weaving approach or the ["migration"](https://tomcat.apache.org/download-migration.cgi) tool than Tomcat has

I would tend to lean toward a hard version move as a 2.0 release myself.

The old javax.servlet users have to be dwindling pretty fast these days. If we continued to maintain a 1.x line for a while that supports javax.servlet and provided 2.x for jakarta, we'd avoid any complex build situation and everyone would still be able to keep working with what they have. We might start dropping some deprecation warnings into the 1.x line at some point.

@kares Are you opposed to this clean break, branching off 1.2.x and releasing a 2.x that uses only jakarta packages going forward?

kares:
> @kares Are you opposed to this clean break, branching off 1.2.x and releasing a 2.x that uses only jakarta packages going forward?

Not at all, just shared my personal opinions, think some folks might still be stuck with older versions here and there due "legacy" (enterprise) reasons. Thus was thinking about ways to support both wout having extra work (such as backports).

I am very happy to see any effort such as this one (a full API rewrite) to keep the jruby-rack project going.

Regarding this PR my only concern is that some methods feel like stubs and should be revisited implementation wise according to the servlet specification.

chadlwilson:
FWIW for [GoCD](https://www.gocd.org/) i'm one of those stuck on javax namespace at the moment due to needing breaking Jetty 12.x upgrades + multiple Spring FW version upgrades which in turn need major Hibernate and Spring Security upgrades - and really only having myself to do it. 😅 But accept that is probably not most people's situation and being on all this EOL stuff is bad for many other reasons.

My main concern about this if this makes it harder to get to a supported Rails version (i.e if no backport is possible) as currently I believe an issue with jruby-rack is preventing me getting to Rails 7+ and 6.2 is also EOL now :(

chadlwilson:
@kares about all I know is still at https://github.com/jruby/jruby-rack/issues/244

I've subsequently seen similar looking session errors with the appraisals on Rails 6.x but yeah, separate to this issue so should discuss elsewhere.

headius:
Another option would be a new named gem that is jakarta-rack or something, but there may be transitive dependency issues getting RubyGems/Bundler to pick that up.

chadlwilson:
Temporarily propagated the removed `javax.servlet.HttpUtils#parseQueryString` into jruby-rack's code and at least the tests are all passing now.

To get this over the line, we probably mainly need to
- be somewhat comfortable `1.2.4` is stable in various folks' setups so we don't have to actively maintain 3 branches.
- get `1.3.0` out (pretty close I think)
- decide what `1.4.0` looks like, other than this change

Ideally we'd also have some stronger Rails regression tests reinstated to bring more confidence in these bigger releases, but that's probably a nice-to-have.

chadlwilson:
Require, no, but because the library is tested with Spring framework they made a decision to require Java 17 to get JEE 9 support thus it cant be tested on lower Java versions anyway. It's not targeting JEE 11 packages.

(As a side note, the Jakarta namespace move is one of the dumbest,.most painful wastes of time in the Java ecosystem for a decade)

ajuckel:
@headius For your "large jruby-rack user that still has to support Java 8 for some time", are they large enough to warrant an attempt at creating a separate Java 8/Jakarta EE 9/Rack ?? version? Or are their needs met with the current release?

ajuckel:
I did some more poking locally, and am caught up with what @chadlwilson was saying. Even if I lower the packages to their EE 9 variants and update the main code to restore Java 8 compatibility, the spring-test dependency is problematic. Spring Test 5.3.39 is using the old javax packages, and 6.0.23 is compiled with Java 17.

There's a non-zero chance that I might poke further at getting the testing side of EE 9+SE 8 resolved but I wouldn't hold your breath. I'm content to agree it would be a non-trivial effort to support EE 9 on Java 8 in jruby-rack.

If I wanted to be more immediately useful, where would folks recommend I direct some effort? I see [this comment](https://github.com/jruby/jruby-rack/pull/266#issuecomment-3170472736) from a few days ago, but I'm not sure where the `1.2.4` or `1.3.0` effort is occurring or what the status is.

chadlwilson:
Thanks for your interest @ajuckel . 

To begin with, I'll dig back and see why I bumped the servlet API further than 5.0. Possibly I thought it didn't do any 'harm' to support the newer API definitions as nothing had been removed, only added.

But the spring test usage is more restrictive. Perhaps we can relook at that. I haven't really dug into its importance outside using it for mock servlet requests etc, but it might be possible to go back to what was done before and copy their mock code into the library, even though I was trying to avoid that maintenance overhead/cruft.

kares:
and before I forget due jruby-rack being both a Java library as well as a Ruby gem, 

assuming the 1.4.0 release is not `javax.servlet` retrofitted but simply a full `jakarta.servlet` rewrite (current PR state) than I believe the **jruby-rack version number should be 2.0** since it's not server "forward" compatible like the 1.x line...

src/main/java/org/jruby/rack/servlet/HttpUtils.java
```diff
@@ -0,0 +1,127 @@
...
+
+package org.jruby.rack.servlet;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.StringTokenizer;
+
+/**
+ * @deprecated As of Java(tm) Servlet API 2.3. These methods were only useful with the default encoding and have been
+ *             moved to the request interfaces. Propagated to jruby-racl from old Javax Servlet API <= 4.0.
```
kares, line 27: small typo: jruby-racl -> jruby-rack

would maybe add that as a separate `@implNote` or `@apiNote`, otherwise good call 👍

src/main/ruby/rack/handler/servlet/default_env.rb
```diff
@@ -279,7 +279,7 @@ def servlet_context
             @servlet_env.servlet_context # ServletRequest#getServletContext()
           else
             if @servlet_env.respond_to?(:context) &&
-                @servlet_env.context.is_a?(javax.servlet.ServletContext)
+                @servlet_env.context.is_a?(jakarta.servlet.ServletContext)
```
kares, line 282: won't work, JRuby only provides some common names at top level such as `java`, `javax`, `org`.

so you'll need to prefix e.g. `Java::jakarta.servlet.ServletContext`.

p.s. `Java::JakartaServlet::ServletContext` (above) is okay.

src/spec/java/org/springframework/mock/web/MockServletContext.java
```diff
@@ -0,0 +1,726 @@
...
+ * @author Rod Johnson
+ * @author Juergen Hoeller
+ * @author Sam Brannen
+ * @since 1.0.2
+ * @see #MockServletContext(ResourceLoader)
+ * @see org.springframework.web.context.support.AnnotationConfigWebApplicationContext
+ * @see org.springframework.web.context.support.XmlWebApplicationContext
+ * @see org.springframework.web.context.support.GenericWebApplicationContext
+ */
+public class MockServletContext implements ServletContext {
```
kares, line 68: 🎉 impressive, a bit more code that anticipated but hopefully it did not take too much of an effort 😅

src/spec/ruby/spec_helper.rb
```diff
@@ -56,7 +56,7 @@ def silence_warnings(&block)
 
   def servlet_30?
     return @@servlet_30 unless @@servlet_30.nil?
-    @@servlet_30 = !! ( Java::javax.servlet.AsyncContext rescue nil )
+    @@servlet_30 = !! ( Java::jakarta.servlet.AsyncContext rescue nil )
```
kares, line 59: whole thing is a bit futile 😉 we can handle this later just fine...

kares (approved):
except one small package rename, this does look great 👍 

do you want to merge into a branch or wait (keep the PR) till 1.3.0 is released on master?

src/main/ruby/rack/handler/servlet/default_env.rb
chadlwilson: Hmm, interesting that no tests fail. Probably because above this line is

```ruby
if @servlet_env.respond_to?(:servlet_context) # @since Servlet 3.0
            @servlet_env.servlet_context # ServletRequest#getServletContext()
else
```

So this clause itself was a <= Servlet 2.5 backward compat. Maybe can be removed?

In an ideal world we'd be able to run tests across multiple servlet API versions to test compatibility more thoroughly...

In any case I should probably change all the references to use the same style, as there are too many combinations here.

chadlwilson:
I suppose it doesn't matter too much which order we do things, but probably wait since we're probably not too far away on 1.3.0?

For this PR, I'll do another sweep for unused mock code, as well as dead servlet <= 5.0 compat code (which definitely won't work after this PR since servlet 5.0 is obviously forward only due to the namespace change, whereas earlier 1.2.x and 1.3.x changes *might* still work with Servlet 2.5 since they didn't tend to remove anything from the impls.)

headius:
> might also need yet another jruby-rack major version change

I guess more interesting to me than the level of effort required is the compatibility impact to existing applications. Can old apps "just work" on Rack 3 if they have up to now been based on earlier versions? What's the real-world impact if we go ahead and do Rack 3 at the same time as Jakarta?

chadlwilson:
Yeah, fair question. I'm no expert, just a wee bit burned by the level of coupling between jruby-rack and rack internals when trying to debug various session issues in later rails versions and having to shuttle between the 3 codebases to figure out why something that used to work no longer does 😅

Given rails itself has been able to support both rack 2.2 and rack 3.0 side-by-side for some time now, it presumably isn't impossible, and aside from splitting rack-session, it doesn't look *too* bad based on https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md

Anyway, sorry for the partially OT comment - can move discussion to https://github.com/jruby/jruby-rack/issues/304 when we get further into it.

PR #270, opened 2025-01-15
Include logger silence from active support

jlahtinen:


chadlwilson:
Hi @jlahtinen - slightly unrelated to this, but which rack version are you using jruby-rack with?

jlahtinen:
> Hi @jlahtinen - slightly unrelated to this, but which rack version are you using jruby-rack with?

rack (2.2.9)

chadlwilson:
Interesting. Do you use with rails?

jlahtinen:
@chadlwilson sorry I did not understand puma reasoning :). Most of the developer time goes running rails app with puma.

We are using rails (6.1.7.7) currently

src/main/ruby/jruby/rack/rails/rails_logger.rb
```diff
@@ -0,0 +1,15 @@
+require 'active_support/logger_silence'
+
+module JRuby
+  module Rack
+    class RailsLogger < JRuby::Rack::Logger
+      include ActiveSupport::LoggerSilence
+
+      def info(*args)
```
kares, line 8: why is the `info` level overriden here?

jlahtinen:
I also think this is not ready to be merged. But I hope we can discuss how silence should be implemented.

I was also thinking that should jruby-rack allow rails to initialize rails default logger somehow when no logger is specified. If I have understood correctly, jruby-rack now sets some jruby-rack specific default logger as rails logger if no other is set. Could this jruby-rack spesific default logger not be the default and let rails set the default like it does when running in e.g. puma? Does this make any sense?

chadlwilson:
Yeah, the project I inherited has some kind of silence related hack as well as a custom slf4j logger. Been meaning to dig into that during the 1.2.x upgrade (since there are.logging changes there too). Currently still stuck on both jruby-rack 1.1 and rails 6.1 due to needing some fixes from here first.

But we're close 👍

jlahtinen:
Could it be an idea to add a configuration option, like `config.x.jruby.use_rails_default_logger = true`? The goal would be to let Rails handle the default logger setup instead of jruby-rack setting its own when not explicitly configured.

The reason for this is that I don't like to set the logger to the default one explicitly. I'm not entirely sure how to do that, and it might change when the Rails version changes.

chadlwilson:
From my perspective that could work, but I am no expert 😅

Sadly we still seems to be a bit short on maintainers here as my PR #271 is stuck (which fixes Rails 7 and reinstates some of the appraisals to sanity check with various Rails versions). 

I am maintaining a fork temporarily (Maven/jar distribution only, not rubygems) at https://github.com/gocd-contrib/jruby-rack if you wanted to give your options a go alongside the appraisals.

kares:
MR is incomplete, someone just needs to figure out the best based way forward (based on supported Rails versions and the `Rails.logger` APIs) and finish it up.

> Sadly we still seem to be a bit short on maintainer time here as my PR https://github.com/jruby/jruby-rack/pull/271 is stuck (which fixes Rails 7 and reinstates some of the appraisals to sanity check with various Rails versions).

Merged in, I do not see any issues with having that in 1.2.x mostly just benefits. Thanks for working on that.

PR #271, opened 2025-01-22
Fix compatibility with Rack 2.x on Rails 7.0+, reinstating Rails appraisal tests

### LINKED ISSUE #168: [to-do] towards **1.2**
**JRuby-Rack** **1.2** is aimed as a cleanup release which might break some binary compatibility with 1.1.x when its worth (a badly designed API or due performance reasons). This should only affect users even when they explicitly depend on JRuby-Rack APIs - the very most users won't be affected at all.

No one seems to be working towards 1.2 much currently. If you're interested in seeing some of these enhancements/features/performance work get in touch or show some support at **[BountySource](https://www.bountysource.com/issues/1446571-to-do-towards-1-2)**
- [ ] performance improvements by going native with crucial .rb parts
  - [x] native JRuby::Rack::Response
  - [x] native JRuby::Rack::Logger
  - [x] native Rack::Handler
- [x] compile against Java 1.6 (likely only JRuby 1.7 will be supported)
- [x] update/compile against Servlet 3.0 (JavaEE 6) at minimum
- [ ] ~~refactor project build - use Mavenfile / JBundler~~
- [ ] when no runtimes specified - default to using a shared one (for all)
- [ ] verify precompiled assets serving (e.g. with Trinidad)
- [ ] support for servlet/filter auto-configuration on JEE 6 scanning ?
- [x] map logger levels 1-to-1
- [x] thread-safe logger.silence by default (with default `JRuby::Rack.logger`) !
- [ ] support `env['rack.errors']`
- [ ] support (default) `env['rack.session']` for all (not just Rails)
  - consider changing **session_options** to a frozen Hash `@env['rack.session.options'] ||= {}`
- [ ] do not read body ... warn on `rewind` by default (review usage in Rack/Rails middleware)
- [ ] alternatives to rack's multi-part (needs rewind and uses temp files for >=128k)
- [ ] default to using the `ServletEnv` for `Rack::Handler::Servlet` (if avoids `rewind`) ?
- [ ] servlet_env file-upload detection ?! ... clearly rack file upload is bloated by using memory.
- [ ] better embedding API - **org.jruby.rack.embed** seems rather confusing (and not that usable)
- [ ] make the listener optional and use the same one for rack as well as rails apps
- [ ] improve `ErrorApp` (HTTP_ACCEPT detection, support for serving 503.json)
- [ ] remove deprecated files e.g. _jruby/rack/errors_
- [x] remove (old) Merb support
- [ ] review booter support setting `jruby.rack.booter = :rails` to force ?
- [ ] possibly release booter, could `RailsBooter.to_app` calls be avoided ?
- [ ] RailsBooter - default to Rails 3/4 - detect 2.3 specifics instead of the other way around
- [x] cleanup old Bundler "ext" + polluted_with_anyio? check in Rack::Response
- [ ] verify Rack::Response send_file extensions are working - include TC support?
- [ ] setup Java/Ruby documentation
- [ ] introduce [Websockets support](https://github.com/jruby/jruby-rack/issues/167)
- [ ] handle [ActionController::Live streaming](https://github.com/jruby/jruby-rack/issues/150)
- [ ] implement [Rack::Async-chronous processing](https://github.com/jruby/jruby-rack/issues/119)
- [ ] review Rails/Sinatra example
- [ ] ~~try out app-engine once again with an example~~

### LINKED ISSUE #244: Session store `@env` is nil with Rails 7

chadlwilson:
This PR makes a few important changes.

- **Requires Rack 2.x** (as bundled), removing old Rack 1.x compatibility code which is very confusing to support (and likely doesn't work)
  - this makes no practical difference *for rails* since rack 2.0 is required from rails 5 anyway.
  -  seems fine to me more widely given rack 1.6 was EOL 5 years ago.
- fixes #244 - using a Java servlet session store under Rails 7.0 was broken
  - Rails 7 accidentally broke backward compat with older store implementations when using flash middleware by needing an `enabled?` function. https://github.com/rails/rails/blob/d4acb7aed7c48c3a884cb0ee52f6de6aa3607f3c/actionpack/lib/action_dispatch/middleware/flash.rb#L71-L72
  - Rack 2.x also made changes to a number of functions jruby rack overrides (nesting `@env` within `@req` for sessions and changing what is passed to functions). https://github.com/rack/rack/commit/c94e22401d4719b4d78378c7b63362cd692f9005
  - The second problem obscured the first problem by breaking the missing method delegation to the Java servlet session.
- Makes a few minor test fixes to the Rails "appraisal" tests to allow them to run with modern JRuby and modern Rails
  - Works around an issue with concurrent-ruby and Rails < 7.1 for tests
  - Corrects visibility on some mock methods
  - Adds workaround for frozen string literal issues with `NameError` handling on older JRuby/Rails versions.
- Adds the appraisals to GitHub Actions as a second job, including Rails `5.0` -> `7.2` for now.
- Updates `README` and `History.md` to reflect 1.2.x releases (assumes this will be in `1.2.3`) 
  - fixes #168 in a sense 

Commits should be logically chunked so can be reviewed independently if necessary. This isn't as scary to review as it looks - almost all of the change chunk #s are from removing and adding appraisal `Gemfile.lock`s 😛

What's NOT done 
- there are some dynamic tests run/selected for specific rails versions based on existence of matching specs and stub directories (e.g `rails41` etc). I have not yet added replacement ones for newer Rails versions or removed the outdated ones yet.
- changing version to 1.3 due to dropping EOL Rails/Rack support - I'll do this if we think it's better though.


chadlwilson:
Would appreciate review/input from @kares and @headius when you folks have time :-)

Appraisals
```diff
@@ -1,43 +1,27 @@
...
-  gem "rails", "~> 4.0.13"
+appraise "rails52" do
+  gem "rails", "~> 5.2.0"
 end
 
-appraise "rails41" do
-  gem "rails", "~> 4.1.16"
+appraise "rails60" do
+  gem "rails", "~> 6.0.0"
 end
```
chadlwilson, line 11: We could consider dropping tests for Rails 5.0 (EOL Apr 2018), 5.2 (EOL Jun 2022) and/or 6.0 (EOL Jun 2023) as well. I kept all versions that worked with JRuby 9.3+.

One reason to be a bit more relaxed with EOLs for JRuby stuff is some of the lag for JRuby itself to catch up with required Ruby versions. Dropping 5.0 is probably pretty reasonable now though.

src/main/ruby/jruby/rack/helpers.rb
```diff
@@ -134,7 +134,7 @@ def resolve_constant(camel_cased_name, context = Object)
             require underscore(camel_cased_name)
             retry
           rescue LoadError => le
-            e.message << " (#{le.message})"
+            e.message = "#{e.message} (#{le.message})"
```
chadlwilson, line 137: Some older Rails/Jruby combinations complain that the `NameError.message` is a frozen string literal when hitting this code; better to avoid assuming I suppose?

src/spec/java/org/jruby/rack/mock/MockHttpSession.java
```diff
@@ -101,7 +101,7 @@ public String getId() {
 		return this.id;
 	}
 
-    void setId(String id) {
+    public void setId(String id) {
```
chadlwilson, line 104: Not visible on modern JRuby otherwise.

src/spec/ruby/action_controller/session/java_servlet_store_spec.rb
```diff
@@ -28,14 +28,11 @@
...
     @app = double "app"
+    @app.stub(:call).and_return [200, {}, ["body"]]
     @env = {"java.servlet_request" => @request, "rack.errors" => $stderr}
     @session_store = ActionController::Session::JavaServletStore.new(@app)
   end
 
-  it "should raise an error if the servlet request is not present" do
-    expect { @session_store.call({}) }.to raise_error(RuntimeError)
-  end
-
```
chadlwilson, line 38: This test was too problematic to get to work reliably across all versions, and seemed to be relying on a lot of other code to get to the expected result from the entry point. Seems to be trying to get into the code below when there is no existing session. I made a call that I didn't think it was sufficiently important after trying to fix it and getting lost as to why it wasn't always raising an error.

https://github.com/jruby/jruby-rack/blob/4fa7813d17216dae1abb77a4fb489a4577d60a96/src/main/ruby/jruby/rack/session_store.rb#L49-L57

src/main/java/org/jruby/rack/servlet/ResponseCapture.java
```diff
@@ -277,17 +277,4 @@ public void setHandledByDefault(boolean handledByDefault) {
...
-    @SuppressWarnings("unchecked")
-    private Collection<String> getHeaderNamesOrNull() {
-        // NOTE: getHeaderNames since Servlet API 3.0 JRuby-Rack 1.1 still supports 2.5
-        try {
-            final Method getHeaderNames = getResponse().getClass().getMethod("getHeaderNames");
-            return (Collection<String>) getHeaderNames.invoke( getResponse() );
-        }
-        catch (NoSuchMethodException e) { return null; }
-        catch (IllegalAccessException e) { return null; }
-        catch (InvocationTargetException e) { return null; }
```
kares, line 290: these kind of changes (dropping servlet major version compatibility) along side with the Rack 1.x compatibility IMO should not be part of a patch version. would justify a 1.3.0 release (regardless how well 1.2.x actually worked with the older Rack/Servlet version).

chadlwilson: The original 1.2 always said it targeted servlet 3.0 though did it not? So I thought the behaviour is no different to what is documented?

I don't really mind. However,.my thinking was that given 1.2..2 did things like bump bundled rack to 3.x (which does not actually work) it seems a bit moot to me?

From my perspective 1.2.2 was kind of a "bad" release which regressed multiple things, (and arguably should have been more of a pre or rc) so I was thinking of 1.2.3 as the first proper release.

But I might be wrong in my impression that 1.2.2 cannot have been working for anyone.

kares: My bad, sorry I am not keeping up with dev here, should really just retire 🙃.
Indeed 1.2 is compiled already against Servlet 3.0 so the clenups make sense. We did not mention Rack in the README so that sounds doable as well, given 1.x is pretty much EOL.

chadlwilson: You possibly still have a valid point, so I am open to opinions of others. My understanding was that `1.2.x` was _basically_ released where you left it :D, and at that time even the unit tests were not running/runnable. 

When I wrote the above and the changelog entries I thought `1.2.0` and `1.2.1` had not been released, but that had been to RubyGems - just not pushed to Maven Central (i install from Jar rather than gem). That might change things a bit - I'll check the delta between these releases.

Subsequent to this I still would like to clean up old rails 2.x and old webserver hacks.

I also realised that I haven't "fully" reinstated the Rails appraisal tests, because there are some dynamic overrides based on presence of test directories matching the appraisal names `rails40`, `rails23` etc, which have actual logic in them to "use rails" in anger. In this PR I have only got the "base" tests running when bootstrapped alongside Rails - not these specific tests which are coupled :-(

kares (approved):

src/main/java/org/jruby/rack/servlet/ResponseCapture.java
kares: Yes, used to be a dual release (RGs and Maven) but with 1.2.x releases only the gem was pushed. Which is a bit unfortunate IMHO. Maybe worth a discussion with current RMs since consuming the artifact is likely embedded Java as well and the release guide still includes the .jar pushing steps...

chadlwilson: I think it’s OK as the publishing to Maven was fixed with 1.2.2 https://mvnrepository.com/artifact/org.jruby.rack/jruby-rack

chadlwilson:
@headius any chance of a review on this one? 🙏

This would get the testing of the 1.2 line into a better (but not perfect) state and bring more confidence to future releases.

chadlwilson:
For anyone interested, I have published a (hopefully temporary) fork of this PR + current `master` available via jitpack.io so I can integrate it into GoCD and get some real "soak" experience with jruby-rack `1.2.x` / Rails 7 / Rack 2.2.11 running at https://build.gocd.org (dogfooding GoCD to build GoCD).

It's available at https://github.com/gocd-contrib/jruby-rack?tab=readme-ov-file#gocd-contrib-fork / `com.github.gocd-contrib:jruby-rack:1.2.3.5`

Intention is to move back to upstream though.

RTrampov:
@chadlwilson I'm using [warbler](https://github.com/jruby/warbler/) to build a .war file for Tomcat, where `jruby-rack` is used as a Tomcat filter and its initialization happens very early. I'm unsure how and if I can require `active_support` earlier.

No other issues with Rails 7.0 or 7.1 so far though.

PR #280, opened 2025-05-30
cleanup support for Rails 2.3 and remove old integration tests

kares:
mostly just removing unused code/specs


kares:
> I must admit that I didn’t understand this stuff, however I wondered whether rather than deleting it, it needed to be adapted to rails 6.1 and 7.x to actually run specific rails tests there

As long as we're talking about  the Rails booter specs then than would depend on whether there's changes to the stuff we're adding/overriding.

There's 2 groups (similar to the removed 2.3 group) left here:

`describe "Rails 3.x", :lib => :stub do` and `describe "Rails 3.1", :lib => [ :stub ] do` so I think the names are not that great anymore (should be pretty much one Rails group or just flat specs with better descriptions).

The idea around `:lib => :stub` is we test stubbed regardless of what actual Rails APIs we run.

> There seemed to be some magic there that looked for these specs if they existed for a given rails version and run them.

There's some like that around in the suite but not in the specs changes here.

chadlwilson:
Sorry, missed your comment here. If you can point me to the branch where you're playing with adding back stubs/tests I can take a look and see if I can work it out, as there was some hacking around to do between Maven and rake to get things like that working on modern JRuby earlier.

In my my main use case, I haven't been able to get Rails 7.1 working (let alone 7.2) and only managed to move to 7.0 recently after getting jruby-rack fixed via the PRs here, so I have less experience with what has changed on later Rails. I'm very surprised 7.2 actually works via your tests, let alone 8.0.

For future it might be a good idea to have the build fail-fast if it can't find stub tests for a given Rails major/minor version to force us to add them and ensure completeness, if indeed they are important (which I think they probably are to validate something "real" even if we are not validating all manner of servlet containers, logging mechanisms and such).

jijiNigeria (changes requested):
Nice

PR #281, opened 2025-05-31
Cleanup Actions workflow and dependabot ignore rules

chadlwilson:
- Cleans up ignore rules to avoid dependabot PRs for likely API-breaking dependencies and reduce PR noise
  - This should automatically clean up all the PRs dependabot created here that otherwise need review  
- Avoids time-consuming double-builds on pushes to PR branches by building only on direct push to `master` and `release`-like branches
- Simplify workflow matrix exclusion rules for JRuby 9.3
- Bumps minor/patch dependency versions for GHA and other dependencies

If you could take a look @kares - just cleaning up after the cherry pick from my earlier JRuby 10 branch. 1.2.x compatible.

Probably doesn't need a release unless this [2.2.16 fix for CGI::Cookie](https://github.com/rack/rack/blob/main/CHANGELOG.md#2216---2025-05-22) is relevant for jruby-rack.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,6 +16,15 @@ updates:
         dependency-type: "production"
       maven-dev-deps:
         dependency-type: "development"
+    ignore:
+      - dependency-name: "org.jruby*"
+        update-types: [ "version-update:semver-major", "version-update:semver-minor" ]
+      - dependency-name: "jakarta*"
+        update-types: [ "version-update:semver-major" ]
+      - dependency-name: "javax*"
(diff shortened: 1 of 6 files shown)


kares:
> Probably doesn't need a release unless this [2.2.16 fix for CGI::Cookie](https://github.com/rack/rack/blob/main/CHANGELOG.md#2216---2025-05-22) is relevant for jruby-rack.

Does not seem like it matters much, but good call to have this in first so we could branch of a `1.2-stable` later...

PR #287, opened 2025-05-31
Bump dependencies in appraisals

chadlwilson:
Synchronize with parent Gemfile.

Would be good to automate these; will need to recall why I didn't ask dependabot to update these at the same time as the parent (possibly it didn't detect them due to the file names?)

Alternative would to not commit the lockfiles and  generate them at runtime from parent, or not use them at all as they are not being run in deployment mode anyway (I think due to some issues with ruby 2.6 compat on JRuby 9.3 that needs to install a different version)


kares:
> Alternative would to not commit the lockfiles and generate them at runtime from parent

Looking into the [integration specs](https://github.com/jruby/jruby-rack/pull/280) they're currently [copied](https://github.com/jruby/jruby-rack/blob/master/src/spec/ruby/jruby/rack/integration_spec.rb#L313-L314) into the Rails 'stub' application, loaded for the tests.

> Would be good to automate these; will need to recall why I didn't ask dependabot to update these at the same time as the parent (possibly it didn't detect them due to the file names?)

Sounds likely that it's only aware of smt like `(G|g)emfile` pattern...

PR #294, opened 2025-07-02
Migrate to publishing via new Sonatype Central publishing portal

chadlwilson:
I believe jruby-rack and jruby share a namespace, and jruby was migrated in https://github.com/jruby/jruby/commit/bea540848c9bb559cfd73f32c2044d8b557f8264 so presume this needs to be done here too.

Have assumed the same `central` ID for creds as JRuby Core is using, in contrast with the earlier `ossrh` credential identifier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ package and push the .jar every time a commit changes a source file).
 
 ## Releasing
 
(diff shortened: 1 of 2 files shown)


kares:
will need to delegate this one, last release I did the existing sonatype setup worked just fine... :shrug:

PR #296, opened 2025-07-06
Fix Rails 7.1 CSRF token support

### LINKED ISSUE #295: Rails 7.1 CSRF / request_forgery_protection broken on jruby-rack session storage
Rails 7.1 changed its CSRF protection approach in https://github.com/rails/rails/pull/44283 to allow storage outside the session. As a side effect of this approach, Rails now stores the token temporarily inside the `request.env` and then hijacks `commit_session` within the Rails `abstract_store` to actually store the CSRF token inside the session just prior to commit. FWIW, this also broke Devise: https://github.com/rails/rails/issues/52244

https://github.com/rails/rails/commit/f2c66ce392016efd49816cd6fdc9f174271a0321#diff-5207bc67aa19cddd5a4997dec3c69fa4ba541750fbd881c6f7e30b27df9ea9ddR70-R73

https://github.com/rails/rails/commit/f2c66ce392016efd49816cd6fdc9f174271a0321#diff-60b77e427ea7ba142faa477fac10b8d0134cede4e35a3b1953c425200fadf1acR433-R435

This hijacking seems to not work with jruby-rack and the CSRF token is lost, causing all subsequent POSTs etc to fail. I believe essentially we have lost "abstract_store compatibility" with this change in Rails.

I have a hack which fixes it, but need to find the "right" way to do this, potentially that is slightly easier to maintain and gets things a bit readier for Rack 3.x.....

```ruby
def commit_session(req, status, headers, body)
  session = req.env[::Rack::RACK_SESSION]
  options = req.env[::Rack::RACK_SESSION_OPTIONS]
  session[:_csrf_token] = req.env["action_controller.csrf_token"] if req.env["action_controller.csrf_token"]
```

@kares are you perhaps able to help make a `1.2-stable` branch I can target for a PR as at https://github.com/jruby/jruby-rack/commit/ff790fb4b2243012ae19615b69a56f7a22743ee1 ?

chadlwilson:
Fixes #295 by replicating logic below from Rails `abstract_store` which propagates the CSRF token from the request environment only during `commit_session`.

https://github.com/rails/rails/commit/f2c66ce392016efd49816cd6fdc9f174271a0321#diff-5207bc67aa19cddd5a4997dec3c69fa4ba541750fbd881c6f7e30b27df9ea9ddR70-R73

https://github.com/rails/rails/commit/f2c66ce392016efd49816cd6fdc9f174271a0321#diff-5f81b06a0e3051b576daee16c960b21e715a6cc26d97d020c546d2fa697c9bc6R345-R348

I couldn't really find an alternate approach that seemed better, e.g by re-using the Rails code directly (either `Request.commit_csrf_token` or `AbstractStore.commit_session`). This is because in the jruby-rack case, the request object that the store sees appears to be a `Rack::Request` whereas the Rails `abstract_store` code expects an `ActionDispatch::Request` so `req.commit_csrf_token` is not defined. I couldn't find a way to cleanly access the `ActionDIspatch::Request` from the `Rack::Request` either.

How this all works earlier is a bit confusing for me, and the current approach is obviously somewhat coupled to the current Rails 7.1/7.2/8.0 behaviour, but it seemed possibly OK given the otherwise deep integration we are dealing with here? The way they have implemented this in Rails doesn't seem great to me due to the need for `AbstractStore` to now understand when to invoke a special CSRF hook, so this has created a bit more coupling for us.

Additional minor technical changes
- Changes sessions to inherit from non-deprecated `::Rack::Session::Abstract::Persisted` to fix warnings
- Avoid unnecessarily overriding `context(env)` and align signatures for overridden functions with Rack 2.2.x for easy of understanding (the only supported Rack version for jruby-rack 1.2)
- Prefer newer impl names from `::Persisted` rather than legacy `::ID` (`delete_session` rather than `destroy_session`, `write_session` rather than `set_session`)
- corrects appraisal lock files for running on older jruby versions

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -2,9 +2,9 @@ name: Java CI
 
 on:
   push:
-    branches: [ "master", "*release*", "*stable*" ]
+    branches: [ "master", "*-stable" ]
   pull_request:
-    branches: [ "master", "*release*", "*stable*" ]
+    branches: [ "master", "*-stable" ]
 
 env:
   # Default versions for canonical release build
--- a/History.md
+++ b/History.md
@@ -4,14 +4,16 @@
 - Drop unnecessary jruby.compat.version and RackConfig.getCompatVersion() API
 - Drop JMS support
 - update (bundled) rack to 2.2.17
+- Fix Rails 7.1 CSRF protection when working with `JavaServletStore` sessions
 
 ## 1.2.4 (UNRELEASED)
 
-- update (bundled) rack to 2.2.16
+- update (bundled) rack to 2.2.17
+- Fix Rails 7.1 CSRF protection when working with `JavaServletStore` sessions
 
 ## 1.2.3
 
-- avoid warnings due usage of `File.exists?`
+- avoid warnings due to usage of `File.exists?`
 - Fix Rails 7.1 compatibility by ensuring active_support is required before railtie
 - Workaround logger require issues with concurrent-ruby 1.3.5 and older Rails versions
 - Workaround NameError frozen string literal issues with JRuby 9.3 and Rails 5.2/6.0
--- a/gemfiles/rails50.gemfile.lock
+++ b/gemfiles/rails50.gemfile.lock
@@ -65,7 +65,7 @@ GEM
     method_source (1.1.0)
     mini_mime (1.1.5)
     minitest (5.25.5)
-    net-imap (0.5.8)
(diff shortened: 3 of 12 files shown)


kares (commented):
:+1: haven't tried or investigated all the changes in Rack but this looks reasonable to me.

also invited you to the jruby-rack team, you should be able to merge these on your own now.

PR #303, opened 2025-07-22
[feat] make logger in Rails include AS::LoggerSilence

kares:
relevant changes in *src/main/ruby/jruby/rack/rails/railtie.rb*

the rest is getting stubs from Rails to be able to test this wout actual Rails

an alternative proposal to: https://github.com/jruby/jruby-rack/pull/270

### DIFF
--- a/src/main/ruby/jruby/rack/rails/railtie.rb
+++ b/src/main/ruby/jruby/rack/rails/railtie.rb
@@ -37,7 +37,11 @@ module JRuby::Rack
         log_formatter = config.log_formatter if config.respond_to?(:log_formatter)
         logger.formatter = log_formatter if log_formatter && logger.respond_to?(:formatter=)
         require 'active_support/tagged_logging' unless defined?(ActiveSupport::TaggedLogging)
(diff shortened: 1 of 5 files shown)


src/main/ruby/jruby/rack/rails/railtie.rb
```diff
@@ -37,7 +37,11 @@ class Railtie < ::Rails::Railtie
         log_formatter = config.log_formatter if config.respond_to?(:log_formatter)
         logger.formatter = log_formatter if log_formatter && logger.respond_to?(:formatter=)
         require 'active_support/tagged_logging' unless defined?(ActiveSupport::TaggedLogging)
-        ActiveSupport::TaggedLogging.new(logger) # returns a logger.clone
+        logger = ActiveSupport::TaggedLogging.new(logger) # returns a logger.clone
+        logger.singleton_class.instance_eval do
+          include ActiveSupport::LoggerSilence if defined?(ActiveSupport::LoggerSilence)
```
kares, line 42: should have just `send(:include)` here instead of eval-ing...

PR #310, opened 2025-08-08
[feat] (re-invent) log4j as a logging backend

kares:


chadlwilson:
Yup, just curious to be honest, since the saga of Java logging seems to be an endless one :-)

I thought perhaps the motivation might be specific to the needs/defaults of particular servlet containers of particular relevance to, or most commonly used with jruby-rack in your experience.

PR #317, opened 2025-08-15
Replace deprecated rspec usage with modern expect usage

chadlwilson:
Some minor reformatting required to make the automated replacements easier.

### DIFF
--- a/src/spec/ruby/action_controller/session/java_servlet_store_spec.rb
+++ b/src/spec/ruby/action_controller/session/java_servlet_store_spec.rb
@@ -11,24 +11,24 @@ describe "ActionController::Session::JavaServletStore" do
 
   before :each do
     @session = double "servlet session"
-    @session.stub(:getId).and_return @session_id = "random-session-id"
-    @session.stub(:getAttribute).and_return nil
-    @session.stub(:getAttributeNames).and_return []
-    @session.stub(:synchronized).and_yield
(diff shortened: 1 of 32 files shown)


kares:
just curious, what did you use to automate the replacements?

chadlwilson:
Good question.

Believe it or not, I threw JetBrains Junie at it which got me 85% of the way there (just with default Claude Sonnet 3.7 underneath) with a bit of coaxing, and then getting annoyed at it breaking some syntax and screwing up some formatting, and not being able to run the tests itself due to `PATH` problems with jruby/java etc. Didn't try Sonnet 4 though. It mainly seemed to construct some dodgy error-prone regexes and then go back and fix syntax after I told it to, so maybe not so different to how I'd have done it, haha.

First time I have used Junie though, I am figuring my way through agent-mode for OSS refactoring, so wanting something easyish like this to have a go with.

Currently I have a Jetbrains Ultimate license and GH Copilot license for OSS usage, but not Jetbrains AI though, so will run out of credits pretty quickly :-)

chadlwilson:
Yeah, I already use GitHub Copilot quite a bit IDE-integrated for auto-completion, back-writing tests and just asking questions of (useful when working in language ecosystems I don't know well) but supervised refactoring like this should be relatively easy stuff for most of these models one would hope.

PR #318, opened 2025-08-19
Remove 1.0/1.1-era API deprecations; Rack < 2.2 code JRuby < 9.3 code

chadlwilson:
More cleanup for 1.3 of things we didn't end up removing for 1.2. 
- old JRuby compatibility code
- some pre Rack 2.2 compatibility code
- jruby-rack API deprecated ourselves.

See the `History.md` changes for more detail on the property/API changes. 

https://github.com/jruby/jruby-rack/pull/318/files#diff-abfa5988643af1b8b2600aac13b273922dbb3372e021bf1d7caadfe7473c9561L9-R16

The subset of these which are pure removal of unnecessary code (not removing public APIs/props) I intend to backport/cherry-pick to `1.2-stable` since they are equally applicable there. (basically the technical changes *not* needing documenting in the changelog/history)


chadlwilson:
These are mainly not contentious since they were deprecated long ago, so will merge this in so I can avoid maintaining too many branches as I have a few things in flight, including a Rack 3.x experiment. I'll revert anything of concern or deal with devlog/servletlog later if necessary.

PR #322, opened 2025-08-23
Reinstate Rails integration test stubs for Rails 6.1, 7.0, 7.1 & 8.0

chadlwilson:
- Generated new Rails dummy apps from scratch for all versions we currently test
- Lifted all of the current initialization tests into the shared examples to run for all Rails
- Added back the chunked rack monkey-patch test that was there for the earlier rails 4.0 era stubs
- Added some gitignores to try and make it easier to generate and add new minimal stub Rails applications
- Added a test that will fail if we forget to add new stub tests that actually boot/initialize Rails for a new version

Looks like
```
rails 8.0
    it should behave like a rails app
      runtime
        initializes pooling when min/max set
        initializes shared (thread-safe) by default
        initializes shared (thread-safe) whem max runtimes is 1
      initialized
        loaded rack ~> 2.2.0
        booted with a servlet logger
        sets up public_path
        disables rack's chunked support (by default)
```

Open question
- [x] the `LoggerSilence` stuff doesn't work on Rails 5.0/5.2 as it doesn't exist there. Should we just drop testing these EOL versions or keep my small change here that makes it apply only to Rails 6.0 onwards?
  - Decided to drop support for Rails 5.0/5.2 and also not test Rails 6.0 now given it's 27 months past EOL.


olleolleolle:
@chadlwilson My take - drop those versions.

chadlwilson:
@olleolleolle thanks for the opinion! Yeah, they are very EOL. I kept them earlier partly because Rails' support policy is quite aggressive, partly because this lib itself was dormant for many years and had issues on Rails 6 in some configurations (with java servlet session store) and to make it easier to reinstate these stub tests so the distance would be smaller between the last version that was actively integration tested here (rails 4.0). But that's largely moot now with this PR since it covers at least framework initialization through to Rails 8.

The only thing that is vaguely of concern is that there are probably some jruby-rack features that are not working properly on modern rails/rack and may also benefit from being able to validate they work incrementally over Rails versions. For example, I think `forward_to` is not working for some reason as I discovered only yesterday (don't use it myself).

Having said that, the Rails 5 tests could live on for the 1.2.x branch for a while since they work fine there.

PR #325, opened 2025-08-30
feat: Rack 3.x and Rails 8.1+ support

### LINKED ISSUE #304: Rack 3.x support
Currently jruby-rack only supports Rack 2.2.x because of some major changes made to rack and splitting out rack-session with 3.0.

At some point soon we'll need to move on, as things will start supporting only Rack 3.

https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md

chadlwilson:
Implements spec compatibility with Rack 3.0 through 3.2.

https://github.com/rack/rack/blob/main/UPGRADE-GUIDE.md

https://github.com/rack/rack/blob/main/SPEC.rdoc

- fully backward compatible with the existing `2.2` support
- adds support for
  - ensuring lower case header names
  - additional SERVER_PROTOCOL env key
  - stricter multi-value handling
  - changes to query parsing semantics (delegated to underlying Rack parser) for the little-used ServletEnv style 
  - Rack 3 spec-complaint `#call` streams (sanity test in the Sinatra example)

TODO list
- [x] "Invalid nested query parsing syntax" support needs review and discussion, especially the inconsistencies in the existing impl depending on
- [x] Need to vendor `rack-session` also? `rackup`? Or rely on Rails/frameworks to bring them in? 
  - Not necessary to vendor, because not everything uses either of them
  - actually not sure we should vendor at all these days since everyone is using bundler, right?
- [x] Check whether removing the `HeaderHash` wrapper for error app has any effect on Rack 2.2
- [x] Remove rewindable support on 3.x since it is no longer a requirement?
  - Won't do this now at least, as things still rely on it, seemingly. (e.g the demo apps)
- [x] Check response header array logic is compatible
  - added much more complete specs
- [x] Investigate whether we supporting response hijacking at all before
  - nope, force disabled so not relevant
- [x] Is `#call` support needed?
  - implemented
- [x] Are the changes to remove optional environment keys worth it, or easier to just keep them?
  - Fine to remove them I think.
- [x] Sanity check with examples and warbler
  - Examples updated to Rack 3.2 and JRuby 10 - all seem fine.


src/spec/ruby/jruby/rack/integration_spec.rb
```diff
@@ -153,10 +153,10 @@
       end
       after(:all) { restore_rails }
 
-      it "loaded rack ~> 2.2.0" do
+      it "loaded default rack version" do
         @runtime = @rack_factory.getApplication.getRuntime
         should_eval_as_not_nil "defined?(Rack.release)"
-        should_eval_as_eql_to "Rack.release.to_s[0, 3]", '2.2'
+        should_eval_as_eql_to "%w(2.2 3.1 3.2).include? Rack.release.to_s[0, 3]", true
```
kares, line 159: IMHO isn't a good test this way, we should maybe duplicate these per each Rails/Rack version tested.

kares: I see now that there's `"rails80" => %w[rack22 rack31 rack32]` so in that sense I guess this is fine...

src/main/ruby/rack/handler/servlet/servlet_env.rb
```diff
@@ -112,9 +112,10 @@ def [](key)
         #
         # @param key the param name
         # @param val the value(s) in a array-like structure
-        # @param hash the Hash to store the name, value pair
-        def store_parameter(key, val, hash)
+        # @param params the Hash to store the name, value pair
+        def store_parameter2(params, key, val, depth = 0)
           # emulating Rack::Utils.parse_nested_query behavior
+          raise ::Rack::Utils::ParamsTooDeepError if depth >= 32
```
kares, line 118: aren't these configurable in Rack itself?

src/main/ruby/rack/handler/servlet/servlet_env.rb
```diff
@@ -112,9 +112,10 @@ def [](key)
         #
         # @param key the param name
         # @param val the value(s) in a array-like structure
-        # @param hash the Hash to store the name, value pair
-        def store_parameter(key, val, hash)
+        # @param params the Hash to store the name, value pair
+        def store_parameter2(params, key, val, depth = 0)
```
kares, line 116: nit-pick: we might want to mark these 2-3 variants as `private` methods

kares (commented):
nice attempt to support both versions, seems maintainable...

src/spec/ruby/jruby/rack/integration_spec.rb
chadlwilson: Yeah it's not good, but I couldn't think of a way to pass in the expected version cleanly without checking loaded gems which is kind of circular....

src/main/ruby/rack/handler/servlet/servlet_env.rb
chadlwilson: It seems just a hard coded security thing :(

chadlwilson: I've reworked both 2.2 and 3.x code now to re-use Rack's query parser utilities which likely didn't exist in this way back when this was written, so the depth safety check is now applying across Rack versions.

PR #450, opened 2026-07-16
[fix] avoid infinite loop in Response#isClientAbortException

### LINKED ISSUE #449: Response#isClientAbortException can lock a CPU core in an endless loop
## Summary

[`Response#isClientAbortException`](https://github.com/jruby/jruby-rack/blob/7ce0039f/src/main/java/org/jruby/rack/ext/Response.java#L619-L630) walks the exception cause chain and will loop for ever on an exception whose cause does not look like a client abort.

  ```java
  protected boolean isClientAbortException(final Exception ioe) {
      String error = ioe.toString();
      if ( error.contains("ClientAbortException") ) return true;
      if ( error.contains("EofException") ) return true;
      while ( true ) {
          if ( error.toLowerCase().contains("broken pipe") ) return true;
          if ( ioe.getCause() == null ) break;
          error = ioe.getCause().getMessage();   // <-- ioe is never reassigned
          if ( error == null ) break;
      }
      return false;
  }
  ```

If the error string does not contail `ClientAbortExceptions` or `EofException`, every iteration re-reads the same `ioe.getCause()` and re-assigns the same message, so unless that message contains "broken pipe" the loop never ends.

## How to reproduce

  ```java
  Exception e = new java.io.IOException(
      new java.util.concurrent.TimeoutException("Idle timeout expired: 30000/30000 ms"));

  isClientAbortException(e); // never returns
  ```

The conditions are: `toString()` contains neither `ClientAbortException` nor `EofException`, and the exception has a non-null cause whose non-null message does not contain "broken pipe".   

## Root cause

see: [`writeBody`](https://github.com/jruby/jruby-rack/blob/7ce0039f/src/main/java/org/jruby/rack/ext/Response.java#L505) calling the `handledAsClientAbort` method:

  ```java
  catch (IOException | RuntimeException e) { if ( ! handledAsClientAbort(e) ) throw e;  }
  ```

so any `IOException` or `RuntimeException` raised while writing a response body will be handled in the loop above. 

Under Jetty, if a streamed response times out because the client is too slow or just has closed the connection, an `IOException` with: `TimeoutException: Idle timeout expired: 30000/30000 ms` can trigger the endless loop. 

This is what happens in ArchivesSpace, which embeds Jetty:

Thread dumps from affected servers showed request threads looping indefinitely inside `isClientAbortException`, repeatedly calling `String.toLowerCase`:

  ```
  "main" #1 cpu=67692.32ms elapsed=70.61s  runnable
        at org.jruby.rack.ext.Response.isClientAbortException(Response.java:627)
        at org.jruby.rack.ext.Response.handledAsClientAbort(Response.java:613)
        at org.jruby.rack.ext.Response.writeBody(Response.java:505)
        at org.jruby.rack.ext.Response.write_body(Response.java:431)
  ```

## Workaround

Setting `jruby.rack.response.swallow_client_abort=false` (the property added in 82bf64cd for #186) makes `handledAsClientAbort` return without ever calling this method, so the loop becomes unreachable. The cost is that genuine client aborts propagate and get logged rather than being swallowed.

A PR with a fix for this is coming up shortly.

thimios:
## Related issue
Fixes #449.

## Summary 

`isClientAbortException` walks the exception cause chain but never advances it, so each iteration re-reads the same cause and re-assigns the same message. See more details in #449.

### The fix

Advance the chain via a separate `cause` variable, and bound the traversal.

The bound is necessary as a cause chain may be cyclic (`a.initCause(b); b.initCause(a)` is possible), which would still hang a `cause = cause.getCause()`.

### DIFF
--- a/src/main/java/org/jruby/rack/ext/Response.java
+++ b/src/main/java/org/jruby/rack/ext/Response.java
@@ -615,15 +615,23 @@ public class Response extends RubyObject implements RackResponse {
         return false;
     }
 
+    /**
+     * How many nested causes {@link #isClientAbortException(Exception)} inspects.
+     * A cause chain is allowed to be cyclic, thus the traversal needs a bound.
+     */
+    private static final int MAX_CAUSE_DEPTH = 20;
+
     // ioe.inspect =~ /(ClientAbortException|EofException|broken pipe)/i
     protected boolean isClientAbortException(final Exception ioe) {
         String error = ioe.toString();
         if ( error.contains("ClientAbortException") ) return true;
         if ( error.contains("EofException") ) return true;
-        while ( true ) {
+        Throwable cause = ioe;
+        for ( int depth = 0; depth < MAX_CAUSE_DEPTH; depth++ ) {
             if ( error.toLowerCase().contains("broken pipe") ) return true;
-            if ( ioe.getCause() == null ) break;
-            error = ioe.getCause().getMessage();
+            cause = cause.getCause();
+            if ( cause == null ) break;
+            error = cause.getMessage();
             if ( error == null ) break;
         }
         return false;
--- a/src/spec/ruby/jruby/rack/response_spec.rb
+++ b/src/spec/ruby/jruby/rack/response_spec.rb
@@ -519,6 +519,49 @@ describe JRuby::Rack::Response do
       end
     end
 
+    it "raises exceptions with a nested cause that do not look like abort exceptions" do
+      servlet_response = org.jruby.rack.mock.fail.FailingHttpServletResponse.new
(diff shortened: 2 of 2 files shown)


chadlwilson:
Thx! This code looks a bit crazy, and I don't understand how it has sat here like this for so long, nor why I have never seen a problem from it myself... 🤔

thimios:
> Thx! This code looks a bit crazy, and I don't understand how it has sat here like this for so long, nor why I have never seen a problem from it myself... 🤔

It is very specific circumstances that would trigger the endless loop which we have been experiencing...

